Issue #8, opened 2023-05-19
Improve test coverage


bobluppes:
I would like to increase the coverage of our unit tests.

A coverage report can be found [here](https://app.codecov.io/github/bobluppes/graaf).

bobluppes:
Closing this issue, as we now have 100% coverage

PR #12, opened 2023-05-21
improve test covr

misarb:
The pull request includes changes that augment the test coverage to `78.38%`, with a focus on the `graph_test` file.
The changes in the pull request encompass the following:

-    Addition of new test cases to cover  the graph module.

- Improvement of existing test cases to include edge cases and handle exceptions appropriately.

I have ensured that all tests pass successfully,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 build/
-log/
+
 
 # IDE
 .vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/
+log/
 
 # IDE
 .vscode/
--- a/test/graaflib/graph_test.cpp
+++ b/test/graaflib/graph_test.cpp
@@ -1,26 +1,30 @@
 #include <graaflib/directed_graph.h>
 #include <graaflib/undirected_graph.h>
 
+
 #include <gtest/gtest.h>
 
 /**
  * Tests which test the common functionality of the graph
  * class go here. Any test specific to a graph specification
  * should go in the respective test files.
-*/
+ */
 
-namespace graaf {
+namespace graaf
+{
 
-    template<typename T>
+    template <typename T>
     struct GraphTest : public testing::Test
     {
         using graph_t = T;
     };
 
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 build/
+log/
 
 # IDE
 .vscode/
-.cache/
\ No newline at end of file
+.cache/
```
bobluppes: Was there a reason to change this file? I would propose to leave it out of the diff

test/graaflib/graph_test.cpp
```diff
@@ -85,5 +116,91 @@ namespace graaf {
...
+        const auto vertex_id_1{graph.add_vertex(10)};
+        const auto vertex_id_2 = graph.vertex_count() + 1;
+
+        ASSERT_THROW({
+            try
+            {
+                // Call the add_edge function with a non-existing vertex ID
+                graph.add_edge(vertex_id_1, vertex_id_2, 100);
+                // If the above line doesn't throw an exception, fail the test
+                FAIL() << "Expected std::out_of_range exception, but no exception was thrown.";
```
bobluppes, line 133: [minor]:
Properly checking this in gtest is quite verbose 😬 

However, if you want you could save a line by removing the `FAIL` call, as the outer most `ASSERT_THROW` would fail the test in case the `add_edge` call does not throw. Same for the other two tests below.

```suggestion
                // Call the add_edge function with a non-existing vertex ID
                graph.add_edge(vertex_id_1, vertex_id_2, 100);
```

bobluppes (commented):
Thanks a lot for working on this, looks great 😃 

I would not change the `.gitignore` in this PR, but once that it adressed I would be happy to merge. Also added a small comment on the exception tests, you can choose whether you want to address it or leave as-is.

Tagging #8 as this addresses part of the issue.

.gitignore
bobluppes: Ah, I see, in that case could you please remove the changes to the `.gitignore` file? Then I can merge the PR.

misarb: yes of course

bobluppes (approved):
Thanks for addressing this, LGTM 👍🏻

Issue #11, opened 2023-05-20
Improve documentation of interfaces


bobluppes:
The documentation [here](https://bobluppes.github.io/graaf/classgraaf_1_1graph.html) is automatically generated using [doxygen](https://www.doxygen.nl/). In order to improve it, we should annotate the interfaces of `graph.h`, `directed_graph.h`, and `undirected_graph.h` with [Javadoc-style](https://www.doxygen.nl/manual/docblocks.html) comments. i.e.: 

```c++
/**
 * Checks whether a vertex with a given ID is contained in the graph. 
 * 
 * @param vertex_id The ID of the vertex we want to check
 * @return A boolean indicating whether the vertex is contained in the graph
 */
[[nodiscard]] bool has_vertex(vertex_id_t vertex_id) const noexcept {...}
```

PR #14, opened 2023-05-25
Added Javadoc-style comments

Relm-Arrowny:
Added Javadoc-style comments for all the public functions and looking at [here](https://bobluppes.github.io/graaf/classgraaf_1_1graph.html), I think doxygen has extract_private set to off so I have not add any comments on any of the private functions.  

I have not change any of the code but I have tested it still compile and pass all test.


bobluppes (approved):
Great, thanks for picking this up! LGTM, thanks for your contribution 🚀 

Tagging #11 as this PR addresses part of the issue.

PR #24, opened 2023-05-29
feat: Vertex degree [class method version]

joweich:
**What it does:**
- Adds the [vertex degree](https://en.wikipedia.org/wiki/Degree_(graph_theory)) for undirected graphs  and the [in-/outdegree](https://en.wikipedia.org/wiki/Indegree) for directed graphs
- Groundwork for future capabilities like detecting [Eulerians cycles and paths](https://en.wikipedia.org/wiki/Eulerian_path)
- Calculating degree for undirected graphs and outdegree for directed ones is essentially just accessing the local neighborhood. However, the indegree needs to loop over the neighborhoods of all vertices

**Food for thought for the future:**

- The performance for indegree is bad since we need to loop over all vertices. At some point, it _might_ be worth trading some memory for performance and add the incoming edges for directed graphs to the adjacency list as well (with an indicator that it's an incoming one)
- Currently, `get_vertex_degree`, `get_vertex_outdegree`, `get_vertex_indegree`, and `get_neighbors` are methods of the graph class or its derivatives. Since they primarly concern a vertex and not a graph, refactoring them into a dedicated vertex (abstract) class could be considered. The user would then inherit their vertex type from this. Tbd how the interface to the adjacency list would need to look like then

### DIFF
--- a/src/graaflib/directed_graph.h
+++ b/src/graaflib/directed_graph.h
@@ -12,6 +12,9 @@ class directed_graph final
   [[nodiscard]] std::size_t get_vertex_outdegree(
       vertex_id_t vertex_id) const noexcept;
 
+  [[nodiscard]] std::size_t get_vertex_indegree(
+      vertex_id_t vertex_id) const noexcept;
+
  private:
   [[nodiscard]] bool do_has_edge(
       vertex_id_t vertex_id_lhs,
--- a/src/graaflib/directed_graph.h
+++ b/src/graaflib/directed_graph.h
@@ -8,6 +8,10 @@ namespace graaf {
 template <typename VERTEX_T, typename EDGE_T>
 class directed_graph final
     : public graph<VERTEX_T, EDGE_T, graph_spec::DIRECTED> {
+ public:
+  [[nodiscard]] std::size_t get_vertex_outdegree(
+      vertex_id_t vertex_id) const noexcept;
+
  private:
   [[nodiscard]] bool do_has_edge(
       vertex_id_t vertex_id_lhs,
--- a/src/graaflib/directed_graph.tpp
+++ b/src/graaflib/directed_graph.tpp
@@ -30,4 +30,10 @@ void directed_graph<VERTEX_T, EDGE_T>::do_remove_edge(
   this->edges_.erase(std::make_pair(vertex_id_lhs, vertex_id_rhs));
 }
 
+template <typename VERTEX_T, typename EDGE_T>
+std::size_t directed_graph<VERTEX_T, EDGE_T>::get_vertex_outdegree(
+    vertex_id_t vertex_id) const noexcept {
+  return (this->get_neighbors(vertex_id)).size();
+}
+
 }  // namespace graaf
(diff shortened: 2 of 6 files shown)


src/graaflib/directed_graph.h
```diff
@@ -8,6 +8,13 @@ namespace graaf {
 template <typename VERTEX_T, typename EDGE_T>
 class directed_graph final
     : public graph<VERTEX_T, EDGE_T, graph_spec::DIRECTED> {
+ public:
+  [[nodiscard]] std::size_t get_vertex_outdegree(
+      vertex_id_t vertex_id) const noexcept;
+
+  [[nodiscard]] std::size_t get_vertex_indegree(
+      vertex_id_t vertex_id) const noexcept;
+
```
bobluppes, line 17: Should it also be possible to compute `get_vertex_degree` for a vertex in a directed graph and get the sum of the in and out vertices?

If we are splitting these methods into free functions we could have the following three functions:

```c++
[[nodiscard]] std::size_t get_vertex_degree(const graaf::graph<V, E, S>&, vertex_id_t);
[[nodiscard]] std::size_t get_vertex_outdegree(const graaf::graph<V, E, S>&, vertex_id_t);
[[nodiscard]] std::size_t get_vertex_indegree(const graaf::graph<V, E, S>&, vertex_id_t);
```

Here the graph parameter has the type of the derived-from class, such that we can accept directed as well as undirected graphs. You could then dispatch to different logic inside each function, i.e.

```c++
template <typename V, typename E, graph_spec SPEC>
std::size_t get_vertex_degree(const graaf::graph<V, E, SPEC> graph, vertex_id_t vertex_id) {
 if constexpr (SPEC == graph_spec::DIRECTED) {
    return get_vertex_outdegree(graph, vertex_id) + get_vertex_indegree(graph, vertex_id);
  } 

  if constexpr (SPEC == graph_spec::UNDIRECTED) {
    // do undirected logic
  }

  // Should never reach this
  std::abort();
}
```

bobluppes (commented):
Thanks for working on this :) The logic looks sound 👍🏻 

From an architecture point of view, I would slightly favor having these degree methods as free functions. Maybe they can go under `graaflib/properties/vertex_properties`. What do you think about this?

Later on we could then implement graph properties such as a predicate `is_regular` to check if each vertex has the same degree for instance.

src/graaflib/directed_graph.h
joweich: > Should it also be possible to compute get_vertex_degree for a vertex in a directed graph and get the sum of the in and out vertices?

I don't think there is a mathematical definition for degree of directed graphs, but if there was one, this would probably be it. So we can definitely go with it.

> If we are splitting these methods into free functions

I don't have a strong opinion here. My rationale was to align them to `get_neighbors`, which is defined as class method and eats a `vertex_id` as well. But I see that they are dealing with a different level of abstraction. Happy to outsource them into free functions. Will do this in a separate PR to enable comparing them side-by-side.

> such that we can accept directed as well as undirected graphs

What would a call of `get_vertex_in/outdegree` on an undirected graph do?

1. Throw exception
2. Return 0 
3. Return the same value as `get_vertex_degree`

I would vote for the third option, since an undirected edge can be seen as both an outgoing and incoming edge.

bobluppes:
FYI, I opened #27, which adds support for header-only installation of the library. When merged, this will introduce conflicts to any changed made under `src/graaflib`.

The conflicts should however be fairly simple to resolve, as the `src` directory was just renamed to `include`. Simply discard any changes made under `src/graaflib` and add them to the respective files under `include/graaflib`.

Feel free to reach out if you encounter any issues 🤖

PR #29, opened 2023-05-30
feat: Vertex degree [free function version]

joweich:
See thread in #24 for details

### DIFF
--- a/include/graaflib/properties/vertex_properties.h
+++ b/include/graaflib/properties/vertex_properties.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include <graaflib/graph.h>
+#include <graaflib/types.h>
+
+#include <concepts>
+
+namespace graaf::properties {
+
+template <typename V, typename E, graph_spec S>
+[[nodiscard]] std::size_t get_vertex_degree(const graaf::graph<V, E, S>& graph,
+                                            vertex_id_t vertex_id);
+
+template <typename V, typename E, graph_spec S>
+[[nodiscard]] std::size_t get_vertex_outdegree(
+    const graaf::graph<V, E, S>& graph, vertex_id_t vertex_id);
+
+template <typename V, typename E, graph_spec S>
+[[nodiscard]] std::size_t get_vertex_indegree(
+    const graaf::graph<V, E, S>& graph, vertex_id_t vertex_id);
+
+}  // namespace graaf::properties
+
+#include "vertex_properties.tpp"
--- a/include/graaflib/properties/vertex_properties.h
+++ b/include/graaflib/properties/vertex_properties.h
@@ -3,8 +3,6 @@
 #include <graaflib/graph.h>
 #include <graaflib/types.h>
 
-#include <concepts>
-
 namespace graaf::properties {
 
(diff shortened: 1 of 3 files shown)


include/graaflib/properties/vertex_properties.tpp
```diff
@@ -0,0 +1,48 @@
...
+std::size_t get_vertex_indegree(const graaf::graph<V, E, S>& graph,
+                                vertex_id_t vertex_id) {
+  if constexpr (S == graph_spec::DIRECTED) {
+    int indegree{0};
+    for (const auto& current_vertex : graph.get_vertices()) {
+      if (graph.get_neighbors(current_vertex.first).contains(vertex_id)) {
+        indegree++;
+      }
+    }
+    return indegree;
```
bobluppes, line 37: Depending on whether we expect graphs to be spart/dense it might also be worth to loop over the edges. I would propose to leave this as-is for now until we have some actual usage though. 

If you are up for it, a slightly more modern C++ experience would be to use `std::algorithm`. However I can also do this in a follow-up.

**Iterating over all vertices**
```c++
// Requires including <algorithm>
return std:ranges::count_if(graph.get_vertices(), [&graph, vertex_id](const vertex_id_to_vertex_t::value_type& kv_pair) {
    const auto& [current_vertex_id, _]{kv_pair};
    return graph.get_neighbors(current_vertex_id).contains(vertex_id);
});
```

**Iterating over all edges**
```c++
// Requires including <algorithm>
return std::ranges::count_if(graph.get_edges(), [vertex_id](const vertex_ids_to_edge_t::value_type& kv_pair) {
    const auto& [edge_id, _]{kv_pair};
    return edge_id.first == vertex_id || edge_id.second == vertex_id;
});
```

bobluppes (approved):
Nice, looks excellent! 

Some smaller comments in case you want to take a look, but can also do that in a follow-up.
Just give me the 👍🏻 and I am happy to merge

include/graaflib/properties/vertex_properties.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+#include <graaflib/graph.h>
+#include <graaflib/types.h>
+
+#include <concepts>
```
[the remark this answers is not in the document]
joweich: Good catch! I might need to get a linter that catches unused imports

include/graaflib/properties/vertex_properties.tpp
joweich: > Depending on whether we expect graphs to be spart/dense it might also be worth to loop over the edges. I would propose to leave this as-is for now until we have some actual usage though.

I think it's hard to come up with a generic approach here. Depending on the graph architecture, the one or the other might be faster.

> If you are up for it, a slightly more modern C++ experience would be to use std::algorithm.

Damn, I feel this wasn't there in C++14. Covered in [50d78ac](https://github.com/bobluppes/graaf/pull/29/commits/50d78ac39e5fdd130a79e7750c834e47c954f256).

bobluppes: Very nice, thanks!

Yes, luckily we have C++20 now 😉

PR #38, opened 2023-06-06
feat: Shortest path for weighted graph

joweich:
This is meant to add Dijkstra's algorithm for calculating the shortest path between two vertices in a weighted graph.


include/graaflib/algorithm/shortest_path.tpp
```diff
@@ -61,6 +61,80 @@ std::optional<GraphPath<int>> get_unweighted_shortest_path(
   }
 }
 
+template <typename V, typename E, graph_spec S>
+std::optional<GraphPath<E>> get_weighted_shortest_path(
+    const graph<V, E, S>& graph, vertex_id_t start_vertex,
+    vertex_id_t end_vertex) {
+  using weight_t = typename graph<V, E, S>::edge_t;
```
bobluppes, line 68: Interestingly enough it seems like clang is confusing the name of the parameter and the type (which are both called `graaf`). Changing the parameter name works, but a slightly more elegant solution is:

```suggestion
    const graph<V, E, S>& graph, vertex_id_t start_vertex,
    vertex_id_t end_vertex) {
  using weight_t = typename decltype(graph)::edge_t;
```

I checked and this should work on both gcc as well as clang.

bobluppes, line 68: I do not get the `no match for ‘operator+’ (operand types are ‘weight_t’ {aka ‘std::shared_ptr<graaf::primitive_numeric_adapter<int> >’} and ‘int’)` error locally, but I suspect this only happens when this new function is called in a test.

I think this could also be related to the current line. `typename(graph)::edge_t` only yields us a `shared_ptr`, see `graph.h` line 38:

```c++
using edge_t =
      std::conditional_t<detail::is_primitive_numeric_v<EDGE_T>,
                         std::shared_ptr<primitive_numeric_adapter<EDGE_T>>,
                         std::shared_ptr<EDGE_T>>;
```

This shared pointer either wraps `primitive_numeric_adapter<EDGE_T>` (if `EDGE_T` is a primitive numeric) or `EDGE_T` otherwise. To get this wrapped type out of the shared pointer we can do `typename decltype(graph)::edge_t::element_type`.

Both `primitive_numeric_adapter<EDGE_T>` as well as `EDGE_T` will derive from `weighted_edge` if we are dealing with a weighted graph. We can get a weight type out of the weighted edge with the trait `weight_t`.

Combining this all:

```c++
using edge_ptr_t = typename decltype(graph)::edge_t;
using wrapped_t = typename edge_ptr_t::element_type;
using weight_t = typename wrapped_t::weight_t;

// Or all combined in one line
using weight_t = typename decltype(graph)::edge_t::element_type::weight_t;
```

bobluppes (commented):
Thanks for working on this! 

I have recently changed the CI to a matrix build such that we ensure the code builds with both gcc as well as clang. Sorry for the inconvenience 😬 

I can do a more in-depth review later, but the overall approach looks good! I mostly looked into the build issues. Let me know if you continue to have issues there.

include/graaflib/algorithm/shortest_path.tpp
joweich: With a minimal test, this proposal fails for me with `error: ‘const graaf::graph<int, graaf::algorithm::my_weighted_edge<int>, graaf::graph_spec::UNDIRECTED>&’ is not a class, struct, or union type`. 
After some googling, adding `std::remove_reference_t` like
```cpp
using edge_ptr_t = typename std::remove_reference_t<decltype(graph)>::edge_t;
```
seems to prevent an error on this exact line. I'm not aware if this change might introduce unintended side effects.
Please note that with the tests provided, the project still doesn't build successfully due to
`from ‘graaf::algorithm::GraphPath<int>’ to ‘std::optional<graaf::algorithm::GraphPath<graaf::algorithm::my_weighted_edge<int> > >` that probably relates to the shared pointer architecture as well. The test is basically your setup for testing weighted edges reduced to integer types.
I pushed my working state in d78e821 for your reference. Help is highly appreciated! Feel free to push to this fork directly if you see an easy fix.

bobluppes:
Hi, sorry for the delay, I finally managed to take a proper look at this.

It turns out the compilation issue was due to the return type of the `get_shortest_path` function. Originally this was defined as `std::optional<GraphPath<E>>`, where `E` is given by `graph<V, E, S>`. However, since this can now be an arbitrary edge class which derives from `weighted_edge`, we want this to be `std::optional<GraphPath<WEIGHT_T>>` instead. See the diff in [be39a50](https://github.com/bobluppes/graaf/pull/38/commits/be39a50e130e95a83fbb0220b0ddc30be358bae0) for more details.

I also made the test suite name globally unique in [a0a60d1](https://github.com/bobluppes/graaf/pull/38/commits/a0a60d1ca8ec40210966a8b558b2755134f4ba64) (apparently gtest requires this).

Finally, I found a bug in the getters of a `const` graph. Calling a getter on the const class resulted in a stack overflow due to an infinite recursion 😅. Quite sneaky and we did not have a test for it. Fixed it on main and merged it into your branch.

Thanks for working on this! The code looks good, if you are up for adding a few tests codecov has flagged a few lines as not covered, but otherwise this looks good to go!

joweich:
@bobluppes This is ready to be merged from my end. The missing coverage for one added line seems to be a bug/configuration issue concerning codecov.
In a follow-up PR, we should think about merging the `reconstruct_path` function in the shortest path logics for weighted and unweighted graphs to reduce code duplication.

bobluppes (approved):
Awesome, thanks for working on this!

Congrats on the third great feature added to Graaf 🥳

PR #42, opened 2023-07-24
DFS cycle detection

Hromz:
- added DFS cycle detection for directed and undirected graph
- added tests

### DIFF
--- a/include/graaflib/algorithm/cycle_detection.h
+++ b/include/graaflib/algorithm/cycle_detection.h
@@ -1,19 +1,16 @@
 #pragma once
 
 #include <graaflib/graph.h>
-#include <graaflib/types.h>
-
-#include <concepts>
 
 namespace graaf::algorithm {
 
 /*
- * @breif Traverses the graph and check for cycles.
+ * @breif Traverses the graph and checks for cycles.
  *
  * @param graph The graph to traverse.
  */
 template <typename V, typename E, graph_spec S>
-bool has_cycle(const graph<V, E, S> &graph);
+[[nodiscard]] bool has_cycle(const graph<V, E, S> &graph);
 
 }  // namespace graaf::algorithm
 
--- a/include/graaflib/algorithm/cycle_detection.h
+++ b/include/graaflib/algorithm/cycle_detection.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <graaflib/graph.h>
+#include <graaflib/types.h>
+
+#include <concepts>
+
+namespace graaf::algorithm {
+
+/*
+ * @breif Traverses the graph and check for cycles.
+ *
+ * @param graph The graph to traverse.
+ */
+template <typename V, typename E, graph_spec S>
+bool has_cycle(const graph<V, E, S> &graph);
+
+}  // namespace graaf::algorithm
+
+#include "cycle_detection.tpp"
--- a/include/graaflib/algorithm/cycle_detection.h
+++ b/include/graaflib/algorithm/cycle_detection.h
@@ -5,7 +5,7 @@
 namespace graaf::algorithm {
 
 /*
- * @breif Traverses the graph and checks for cycles.
+ * @brief Traverses the graph and checks for cycles.
  *
  * @param graph The graph to traverse.
  */
--- a/include/graaflib/algorithm/cycle_detection.tpp
+++ b/include/graaflib/algorithm/cycle_detection.tpp
@@ -1,31 +1,32 @@
 #pragma once
 
+#include <graaflib/types.h>
+
 #include <algorithm>
 #include <unordered_map>
 
 namespace graaf::algorithm {
 
+enum vertex_color { UNVISITED, VISITED, NO_CYCLE };
 namespace detail {
-
-enum vertex_color{WHITE, GREY, BLACK};
 
 template <typename V, typename E, graph_spec S>
 bool do_dfs_directed(const graph<V, E, S>& graph,
-                     std::unordered_map<vertex_id_t, int>& colored_vertices,
+                     std::unordered_map<vertex_id_t, vertex_color>& colored_vertices,
                      vertex_id_t current) {
-  colored_vertices[current] = vertex_color::GREY;
+  colored_vertices[current] = vertex_color::VISITED;
 
   for (const auto& neighbour_vertex : graph.get_neighbors(current)) {
-    if (colored_vertices[neighbour_vertex] == vertex_color::WHITE) {
+    if (colored_vertices[neighbour_vertex] == vertex_color::UNVISITED) {
         if (do_dfs_directed(graph, colored_vertices, neighbour_vertex))
             return true;
     } 
-    else if (colored_vertices[neighbour_vertex] == vertex_color::GREY) {
+    else if (colored_vertices[neighbour_vertex] == vertex_color::VISITED) {
       return true;
     }
   }
 
-  colored_vertices[current] = vertex_color::BLACK;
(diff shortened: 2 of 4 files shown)


include/graaflib/algorithm/cycle_detection.h
```diff
@@ -0,0 +1,20 @@
...
+
+namespace graaf::algorithm {
+
+/*
+ * @breif Traverses the graph and check for cycles.
+ *
+ * @param graph The graph to traverse.
+ */
+template <typename V, typename E, graph_spec S>
+bool has_cycle(const graph<V, E, S> &graph);
```
bobluppes, line 16: [very minor]: I would prefer to mark predicates such as these with `[[nodiscard]]`

include/graaflib/algorithm/cycle_detection.tpp
```diff
@@ -0,0 +1,96 @@
...
+                                      vertex.first, parent_vertices[vertex.first])) {
+                         return true;
+            }
+        }
+  }
+
+  return false;
+
+  // We should never reach this
+  std::abort();
```
bobluppes, line 93:
```suggestion
  return false;
```

include/graaflib/algorithm/cycle_detection.tpp
```diff
@@ -0,0 +1,96 @@
...
+            if (!colored_vertices.contains(vertex.first) &&
+                detail::do_dfs_directed(graph, colored_vertices, vertex.first)) {
+                        return true;
+            }
+        }
+
+    return false;
+  }
+
+  if (graph.is_undirected()) {
```
bobluppes, line 77: [minor]: in the case of an undirected graph, we can return early if the number of edges is equal to or greater than the number of vertices. What do you think of adding this here?
```suggestion
  if (graph.is_undirected()) {
       if (graph.edge_count() >= graph.vertex_count()) {
         return true;
       }
```

include/graaflib/algorithm/cycle_detection.tpp
```diff
@@ -0,0 +1,96 @@
+#pragma once
+
+#include <algorithm>
+#include <unordered_map>
+
+namespace graaf::algorithm {
+
+namespace detail {
+
+enum vertex_color{WHITE, GREY, BLACK};
```
bobluppes, line 10: [minor]: Could you please provide a short comment to clarify these enum values? Or maybe change the names to something like:
```suggestion

enum vertex_color{UNVISITED, VISITED, NO_CYCLE};
```

include/graaflib/algorithm/cycle_detection.tpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+}  // namespace detail
+
+template <typename V, typename E, graph_spec S>
+bool has_cycle(const graph<V, E, S>& graph) {
+
+
+  if (graph.is_directed()) {
+        std::unordered_map<vertex_id_t, int> colored_vertices{};
```
bobluppes, line 65:
```suggestion
  if (graph.is_directed()) {
        std::unordered_map<vertex_id_t, vertex_color> colored_vertices{};
```

bobluppes: I would then also make `vertex_color` an `enum class` rather than a plain `enum` now that we do not require the implicit conversion to `int` anymore.

bobluppes (approved):
Thanks for working on this, looks very good!
Only some minor comments, let me know if you want to address this before merging.

include/graaflib/algorithm/cycle_detection.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <graaflib/graph.h>
+#include <graaflib/types.h>
+
+#include <concepts>
```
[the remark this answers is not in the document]
Hromz: Alright ill do

include/graaflib/algorithm/cycle_detection.tpp
Hromz: Agree, wanted to do in first place but changed my mind in the end :)

include/graaflib/algorithm/cycle_detection.tpp
Hromz: Hmm, ill add comments when for graph coloring

Hromz:
Hey @bobluppes, thanks for your comments! Ill change according to them

include/graaflib/algorithm/cycle_detection.h
Hromz: Ok

include/graaflib/algorithm/cycle_detection.tpp
Hromz: Good suggestion, since it's undirected graph and we cannot add edge with vertices that out of range this one will be correct!

include/graaflib/algorithm/cycle_detection.tpp
```diff
@@ -0,0 +1,102 @@
...
+#include <graaflib/types.h>
+
+#include <algorithm>
+#include <unordered_map>
+
+namespace graaf::algorithm {
+
+enum vertex_color { UNVISITED, VISITED, NO_CYCLE };
+namespace detail {
+
```
bobluppes, line 12: I would still make this an `enum class` rather than an `enum`. Then the values do not pollute the namespace it is in. I would then also move it inside the detail namespace to keep the scope as small as possible.
```suggestion

namespace detail {

enum class vertex_color { UNVISITED, VISITED, NO_CYCLE };

```

include/graaflib/algorithm/cycle_detection.tpp
```diff
@@ -0,0 +1,102 @@
...
+}  // namespace detail
+
+template <typename V, typename E, graph_spec S>
+bool has_cycle(const graph<V, E, S>& graph) {
+
+  if (graph.is_directed()) {
+        std::unordered_map<vertex_id_t, vertex_color> colored_vertices{};
+
+        for (const auto& vertex : graph.get_vertices()) {
+            if (colored_vertices[vertex.first] == UNVISITED &&
```
bobluppes, line 68: This would then be:
```suggestion
        std::unordered_map<vertex_id_t, detail::vertex_color> colored_vertices{};

        for (const auto& vertex : graph.get_vertices()) {
            using enum detail::vertex_color;
            if (colored_vertices[vertex.first] == UNVISITED &&
```

include/graaflib/algorithm/cycle_detection.tpp
```diff
@@ -0,0 +1,102 @@
...
+                         return true;
+            }
+        }
+  }
+
+  return false;
+
+  // We should never reach this
+  std::abort();
+  return false;
```
bobluppes, line 99: I still think we can simplify this a bit, as we could never reach any line after the first return statement:
```suggestion

  return false;
```

include/graaflib/algorithm/cycle_detection.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include <graaflib/graph.h>
+
+namespace graaf::algorithm {
+
+/*
+ * @breif Traverses the graph and checks for cycles.
+ *
+ * @param graph The graph to traverse.
+ */
```
bobluppes, line 11: [very minor]: last typo
```suggestion
/*
 * @brief Traverses the graph and checks for cycles.
 *
 * @param graph The graph to traverse.
 */
```

bobluppes (approved):
Awesome, thanks for finishing this! Just some very minor comments from which I will submit the suggestions before merging, so no need to take action on your end.

Congrats on the first PR in Graaf 🥳

PR #46, opened 2023-07-30
tooling: Benchmark visualization script

bobluppes:
This PR adds a python script under `tools/`, which can be used to visualize google benchmark data. An example plot can be found below.

Not very familiar with python, so any improvement suggestions are more than welcome.

vertex

### DIFF
--- a/tools/benchmark_visualization.py
+++ b/tools/benchmark_visualization.py
@@ -1,0 +1,46 @@
+import pandas as pd
+import matplotlib.pyplot as plt
+from pathlib import Path
+import os
+
+benchmark_dir = Path("../build")
+benchmark_filter = "bm_"
+google_benchmark_skiprows = 10
+
+plt.figure()
+
+for benchmark_results in benchmark_dir.glob("*.csv"):
+    filename = os.path.basename(benchmark_results)
+    benchmarks_df = pd.read_csv(benchmark_results, skiprows=google_benchmark_skiprows)
+
+    # The benchmark name and variable are recorded as 'bm_name/variable'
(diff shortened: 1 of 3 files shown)


tools/benchmark_visualization.py
```diff
@@ -0,0 +1,57 @@
...
+import os
+
+"""
+This script can be used to visualize benchmark results. The script assumes multiple benchmark 
+results in csv files under the build directory. These results can be generated with the 
+Graaf_perf executable:
+
+build/perf/Graaf_perf --benchmark_out=benchmark_results.csv --benchmark_out_format=csv
+"""
+
```
joweich, line 13: It's an idiomatic [best practice](https://stackoverflow.com/a/419185) to wrap your executed code into `if __name__ == '__main__' ` to avoid unintended side effects. In you're case, it won't make a difference, but it's a good practice nevertheless

joweich (commented):
@bobluppes Bear with me, I've been very picky with my comments. In general, I think there are some low hanging opportunities to make the code more python-idiomatic. Happy to give it a shot and add my proposal to this PR!

joweich (approved):
There's definitely still some potential for additional checks and capabilities, esp. automatic inference of `skiprows`, but I think it's fine for now.

bobluppes:
> There's definitely still some potential for additional checks and capabilities, esp. automatic inference of `skiprows`, but I think it's fine for now.

Thanks for improving this, it looks much more pythonic now 🐍 

Added a very minor change such that we can now glob subdirectories as I just changed to the following build layout
```
build
 |_ Debug
 |_ Release
```

PR #47, opened 2023-08-01
Example transport path

Hromz:
First simple example


docs/docs/examples/example-basics/transport-example.md
```diff
@@ -0,0 +1,118 @@
...
+void print_visited_vertices(const graaf::undirected_graph<station, road>& graph,
+                            seen_vertices_t& seen,
+                            const std::string& filepath) {
+...
+}
+```
+
+Creating a vertex callback structure that will be passed as an argument in the graph traverse function
+The function is needed in order to be called inside the traverse function; see graph.tpp file lines 15 and 45 for context.
+
```
bobluppes, line 81: [minor]: In general I would avoid directly referencing line numbers in source files as this is subjective to change quite often.
```suggestion

Creating a vertex callback structure that will be passed as an argument in the graph traverse function
The function is needed in order to be called inside the traverse function; see graph.tpp for context.

```

docs/docs/examples/example-basics/transport-example.md
```diff
@@ -0,0 +1,118 @@
...
+    seen_vertices_t& seen_vertices;
+
+    record_vertex_callback(seen_vertices_t& seen_vertices)
+        : seen_vertices{seen_vertices} {}
+
+    void operator()(const graaf::vertex_id_t vertex) const {
+        seen_vertices.insert(vertex);
+    }
+};
+```
```
bobluppes, line 94: I think the results from your BFS graph traversal are missing here. For instance, the `example_traversed_graph_BFS.png` is not added in this markdown file.

examples/transport/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+struct station {
+  std::string name{};
+  double gaz_capacity{};
+};
+
+struct road {
+  int length{};
+  int get_weight() { return length; }
+};
```
bobluppes, line 16: [minor]: I know this code works currently, but ideally the road struct should derive from `weighted_edge`  and override the `get_weight` method. This is a design flaw on the weighted edge design, which I plan to address in a separate PR. 

For now, I would do the following;

```suggestion
struct road : public weighted_edge<int> {
  int length{};
  [[nodiscard]] int get_weight() const noexcept override { return length; }
};
```

bobluppes (changes requested):
Thanks for working on this, looks quite good already! 

I like your example graph, as it lends itself well for demonstrating the selected algorithms. I was thinking if we could change some of the vertex/edge data to make the example more realistic, let me know what you think about this.

bobluppes:
FYI, I merged quite a few breaking changes to the `graaf` interface on main. I pushed the changes to your branch, so you should be able to continue (after pulling your branch) without any merge conflicts. Please feel free to reach out if you encounter any issues.

Hromz:
Hey @bobluppes! Can you take a look? For some reason compile and test doesnt pass but it works fine on my machine. I've checked the error and it's saying that cannot find breadth traversal.

bobluppes (approved):
Thanks for working on this, looks great! I really like the example with the station names 🚀  

The CI issue turned out to be a conflict after merging main (which the CI does before running the workflows). On main I implemented our idea of changed the BFS/DFS callback to record traversed edges rather than vertices. I merged main on your branch and adjusted the callback accordingly.

Issue #63, opened 2023-08-07
[FEAT] report visited edges in BFS/DFS


bobluppes:
### Why is this important?
Most use cases that make use of `breadth_first_traverse` or `depth_first_traverse` are actually interested in finding the traversed edges, rather than the traversed nodes.

### Describe the solution you'd like
Instead of the BFS/DFS callback having the signature `void(vertex_id_t)`, I would propose to change the signature of the callback to `void(vertex_id_t, vertex_id_t)` and to call it for every visited edge.

### Additional context
We should probably do something similar for the `shortest_path` algorithms.

### Checklist
Please check the following before submitting the issue:

- [x] I have searched for similar feature requests in the issues.
- [x] This feature is not already implemented in the library.
- [x] I have provided a clear and concise description of the feature request.
- [x] I have explained why this feature is important and how it benefits the library and users.
- [x] I am willing to contribute to the development of this feature (if applicable).

Please note that feature requests are subject to review and may or may not be implemented in the library.

teito-dev:
Hi, I do want to traverse nodes instead of edges, what do i do in my case ? 
The previous implementation was perfect for me, cant we retain both edge callback and vertex callback options for dfs and bfs search ?

For now, i've implemented my own DFS algo for traversal.

Also, Thanks a lot for making this lib ! , you saved me from Boost :)

I'd be happy to code it and send a pull request to you, based on your Aug 8, 2023 commit which had the node callback.
If you think its a good idea to have both edge callback and vertex callback options.

In my particular usecase, I'm working with a in-house custom 2D Video Renderer, and I'm using your graphlib to represent the Scenegraph of my videos, that I need to render. I'm using DFS and traversing and calling the render function on each node, to recursively render out each animated frame in a video.

PR #64, opened 2023-08-07
breaking-change: Report traversed edges instead of vertices

bobluppes:
This PR addresses #63.

Instead of reporting the traversed vertices in `breadth_first_traverse` and `depth_first_traverse`, we report the traversed edges.

In a follow up I would make a similar change to the shortest path algorithms, as a path given by a range of edges is more usable. See our current shortest path example.


bobluppes:
FYI @Hromz

Issue #55, opened 2023-08-06
[ALGO] Bellman-Ford Shortest Path


bobluppes:
# Bellman-Ford Shortest Path
The Bellman-Ford algorithm computes the shortest path between a source vertex and all other vertices in the graph. Compared to Dijkstra, it can handle negative edge weights. See the [wikipedia](https://en.wikipedia.org/wiki/Bellman%E2%80%93Ford_algorithm) entry for more details.

## Syntax
The algorithm should have the following syntax:

```c++
template <typename V, typename E, graph_type T>
[[nodiscard]] return_type your_algorithm_name(const graph<V, E, T>& graph);

/**
 * Find the shortest paths from a source vertex to all other vertices using
 * the Bellman-Ford algorithm.
 *
 * @tparam V The vertex type of the graph.
 * @tparam E The edge type of the graph.
 * @tparam T The graph specialization (directed or undirected).
 * @tparam WEIGHT_T The type of weight associated with the edges.
 * @param graph The graph in which to find the shortest paths.
 * @param start_vertex The source vertex for the shortest paths.
 * @return A map of target vertex IDs to shortest path structures. Each
 *         value contains a graph_path object representing the
 *         shortest path from the source vertex to the respective vertex.
 *         If a vertex is unreachable from the source, its entry will be
 *         absent from the map.
 */
template <typename V, typename E, graph_spec T, typename WEIGHT_T = decltype(get_weight(std::declval<E>()))>
std::unordered_map<vertex_id_t, graph_path<WEIGHT_T>>
bellman_ford_shortest_paths(const graph<V, E, T>& graph, vertex_id_t start_vertex);

```

This should live in the `graaf::algorithm` namespace under `include/graaflib/algorithm/shortest_path.h`.

## Definition of Done
This issue is done when:

- [ ] The algorithm is implemented
- [ ] The new function has a javadoc-style comment explaining the interface
- [ ] Appropriate tests are added under `test/graaflib/algorithm/shortest_path_test.cpp`
- [ ] A test coverage of at least 95% is reached
- [ ] A documentation entry is added under `docs/docs/algorithms` under the appropriate category
  - Just adding a short description and the algorithm syntax here is fine
  - See the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation) on how to build the documentation locally
- [ ] The algorithm is added to the list of algorithms in `README.md`

ndcroos:
Hi, I'd like to work on this issue.

[...]

PR #68, opened 2023-08-12
Bellman-Ford without negative weight cycle detection

ndcroos:
See https://github.com/bobluppes/graaf/issues/55
This PR attempts to covers all subtasks, except test coverage of at least 95%.
Documentation is a short description for now.
Test cases are now taken from those for Dijkstra's shortest paths method.


bobluppes:
Hi @ndcroos, thank you for your contribution!

 Looking at the CI, it seems the `clang-format` and `test` workflows are failing. The format workflow could be fixed by running `clang-format` on your branch (more details on the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#formatting)).
 
 For the tests, it seems that some of the new tests either have a failing assert or a segfault. Maybe you could already take a look at this.

include/graaflib/algorithm/shortest_path.tpp
```diff
@@ -153,4 +153,45 @@ dijkstra_shortest_paths(const graph<V, E, T>& graph,
...
+
+  std::unordered_map<vertex_id_t, graph_path<WEIGHT_T>> shortest_paths;
+
+  // Initialize the shortest path distances from the starting vertex to "infinity" for all vertices
+  for (const auto& vertex : graph.get_vertices()) {
+    shortest_paths[vertex.first] = {
+      {vertex.first},
+      std::numeric_limits<WEIGHT_T>::max()
+    };
+  }
```
bobluppes, line 168: [very minor]: You could use structured binding here to get rid of the `pair` type. This makes the variable names slightly more readable:
```suggestion
  for (const auto& [vertex_id, _] : graph.get_vertices()) {
    shortest_paths[vertex_id] = {
      {vertex_id},
      std::numeric_limits<WEIGHT_T>::max()
    };
  }
```

include/graaflib/algorithm/shortest_path.tpp
```diff
@@ -153,4 +153,45 @@ dijkstra_shortest_paths(const graph<V, E, T>& graph,
...
+    {start_vertex}, 
+    0
+  };
+
+  // Relax edges for |V| - 1 iterations
+  for (std::size_t i = 1; i < graph.vertex_count(); ++i) {
+    for (const auto& edge : graph.get_edges()) {
+      const auto [u, v]{edge.first};
+      WEIGHT_T weight = get_weight(edge);
+
```
bobluppes, line 181: Here, `graph.get_edges()` returns an `unordered_map` of `edge_id_t` to `edge_t`. Iterating over it means that the type of `edge` here is `std::pair<edge_id_t, edge_t>`. 

As a result, the call `get_weight(edge)` is not able to extract the actual weight of the edge, but instead will always return a unit weight. Instead, it should be called with the actual `edge_t:

```c++
WEIGHT_T weight = get_weight(edge.second);
```

To make this more clear in the code, I would also suggest to use structured binding:
```suggestion
    for (const auto& [edge_id, edge] : graph.get_edges()) {
      const auto [u, v]{edge_id};
      WEIGHT_T weight = get_weight(edge);

```

I assume this might also be the reason why the tests fail.

bobluppes (commented):
Looks very good already, thanks for working on this!

Some minor comments, and an issue with the handling of the edge weights. I just checked locally and I think adjusting the weight handling should fix the failing tests.

bobluppes (approved):
Looks great, thanks for contributing 🚀 
I created #76 to keep track of the negative weight cycle detection.

PR #71, opened 2023-08-13
Enhancement: Catch negative edge weights in tree-version dijkstra

joweich:
Complementary to #67.

Note that I again used `std::format`, which I see to be removed in the whole codebase in a follow-up PR.


bobluppes (approved):
Nice, thank you!

I have created #72 to keep track of removing `fmt`. Feel free to pick it up if you want to, otherwise I can take a look next week.

PR #73, opened 2023-08-13
docs: add documentation for BFS (#56)

MichaeINeumann:
Hello, I made the documentation for BFS.
Please Check the ## Syntax section. I understood not exactly what I did here.
Also I'm not sure with the content at ### BFS is particularly useful for: , because I not used this algorithms in praxis.
I have only experience with dijkstra in robot programming.

### DIFF
--- a/docs/docs/algorithms/traversal/breadth-first-search.md
+++ b/docs/docs/algorithms/traversal/breadth-first-search.md
@@ -3,3 +3,97 @@ sidebar_position: 1
 ---
 
 # Breadth First Search (BFS)
+
+## Breadth First Search (BFS) Algorithm
+
+Breadth First Search (BFS) is a fundamental graph traversal algorithm used to explore and analyze graphs, be they directed or undirected. It operates on the principle of visiting nodes in layers, starting from a given source node and gradually expanding outward to neighboring nodes at increasing distances. BFS ensures that all nodes at a particular distance from the source are visited before moving on to nodes at a greater distance. This process continues until all reachable nodes have been visited, forming a breadth-first exploration of the graph.
+
+The BFS algorithm can be succinctly described using the following steps:
+
+1. Begin by selecting a source node as the starting point of the traversal and enqueue it in a queue data structure.
+
+2. While the queue is not empty, repeat the following steps:
+   a. Dequeue a node from the front of the queue.
+   b. Process the dequeued node, which may involve examining its attributes, marking it as visited, or performing other relevant operations.
+   c. Enqueue all unvisited neighbors of the dequeued node into the queue.
+
+3. Continue this process until the queue becomes empty, indicating that all reachable nodes have been visited.
+
+
+### BFS is particularly useful for:
+
+- P2P - Find neighbor nodes:
+  - Finds all neighbors, and then all neighbors of these neighbors.
+
+- Search Engine Crawler:
+  - Helps in systematically crawling web pages, exploring links layer by layer.
+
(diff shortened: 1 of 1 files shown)


bobluppes:
Thanks for your contribution 🚀 
It looks very detailed, I will take a closer look tomorrow. Tagging #56

docs/docs/algorithms/traversal/breadth-first-search.md
```diff
@@ -2,4 +2,98 @@
...
+- `graph`: The graph to traverse. This parameter represents the graph data structure on which the traversal will be performed.
+```graph<int, int> sample_graph; ```
+
+- `start_vertex`: Vertex id where the traversal should be started. This parameter specifies the initial vertex from which the traversal begins.
+
+- `edge_callback`: A callback function that is called for each traversed edge. It should be invocable with an `edge_id_t` object, representing an edge in the graph.
+
+- `search_termination_strategy`: A unary predicate that indicates whether the traversal should continue or not. The traversal continues while this predicate returns `false`. This parameter is optional and defaults to a predefined search termination strategy.
+
+- Return Type: The provided code does not explicitly return a value. The traversal is performed by visiting vertices and edges in the graph based on the specified parameters.
```
bobluppes, line 99: The current proposal for the `Syntax` section is to have a codeblock displaying the signature of the function without the java doc and without examples. As such, I would slighly alter this section to something like the suggestion below.

Let me know what you think. Also feel free to accept/reject the suggestion, as I would be happy to merge it either way.

````suggestion
## Syntax

The bfs_termination_strategy returns true when a certain condition is met, causing to terminate. The bfs_edge_callback is a function that is used as a callback during the BFS traversal to perform some action whenever an edge is traversed.

```cpp
template <
    typename V, typename E, graph_type T,
    typename EDGE_CALLBACK_T = detail::noop_callback,
    typename SEARCH_TERMINATION_STRATEGY_T = detail::exhaustive_search_strategy>
  requires std::invocable<EDGE_CALLBACK_T &, edge_id_t &> &&
           std::is_invocable_r_v<bool, SEARCH_TERMINATION_STRATEGY_T &,
                                 vertex_id_t>
void breadth_first_traverse(
    const graph<V, E, T> &graph, vertex_id_t start_vertex,
    const EDGE_CALLBACK_T &edge_callback,
    const SEARCH_TERMINATION_STRATEGY_T &search_termination_strategy =
        SEARCH_TERMINATION_STRATEGY_T{});
```

### Explanation of Parameters:

- **graph**: The graph to traverse. This parameter represents the graph data structure on which the traversal will be performed.
- **start_vertex**: Vertex id where the traversal should be started. This parameter specifies the initial vertex from which the traversal begins.
- **edge_callback**: A callback function that is called for each traversed edge. It should be invocable with an `edge_id_t` object, representing an edge in the graph.
- **search_termination_strategy**: A unary predicate that indicates whether the traversal should continue or not. The traversal continues while this predicate returns `false`. This parameter is optional and defaults to a predefined search termination strategy, which traverses the graph exhaustively.
- **return**: The provided code does not explicitly return a value. The traversal is performed by visiting vertices and edges in the graph based on the specified parameters.
````

bobluppes (approved):
Hi @MichaeINeumann thank you for the contribution! The section looks very detailed and provides very good examples of BFS use cases.

Only one comment on the `Syntax` section, let me know what you think and we can merge it 🚀

bobluppes:
Thanks, LGTM 👍🏻

PR #75, opened 2023-08-13
Remove dependency to fmt::format in library

### LINKED ISSUE #72: Remove usage of fmt in library
In order for clients to include the library header-only, we should not have any external dependencies. As such, I would like to remove the usage of `fmt` in the library. We can keep it in the tests and examples.

Ideally we would replace it by `std::format`, but I remember GCC and Clang having some issues with their implementations in the std. 

The goal of this issue is to investigate whether we can use `std::format` instead of `fmt`, and replace it if possible. If this is not (yet) possible, we should do the string concatenation manually.

joweich:
Will eventually close #72.

### DIFF
--- a/include/graaflib/algorithm/shortest_path.tpp
+++ b/include/graaflib/algorithm/shortest_path.tpp
@@ -3,6 +3,7 @@
 #include <graaflib/algorithm/graph_traversal.h>
 
 #include <algorithm>
+#include <format>
 #include <optional>
 #include <queue>
 #include <unordered_map>
@@ -100,7 +101,7 @@ std::optional<graph_path<WEIGHT_T>> dijkstra_shortest_path(
       WEIGHT_T edge_weight = get_weight(graph.get_edge(current.id, neighbor));
 
       if (edge_weight < 0) {
-        throw std::invalid_argument{fmt::format(
+        throw std::invalid_argument{std::format(
             "Negative edge weight [{}] between vertices [{}] -> [{}].",
             edge_weight, current.id, neighbor)};
       }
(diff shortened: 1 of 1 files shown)


joweich:
@bobluppes The CI's build results are expected as it uses gcc v11.4, which does not support `std::format`. If it would not be cancelled, the clang build _should_ be fine as clang 14 supports `std::format`.
While [gcc v13 supports it](https://stackoverflow.com/a/69156536), it looks like gcc v13 isn't yet available directly for major operating systems (e.g. [Ubuntu LTS <= 22.04](https://stackoverflow.com/questions/76168157/installing-gcc-13-on-ubuntu-22-04), which I can confirm for my PopOS 22.04).
My proposal would be scrapping all formatting libs and do old school string concatenation.

MichaeINeumann:
Maybe that is a possible solution: @joweich That is what you mean?

```
    for (const auto& neighbor : graph.get_neighbors(current.id)) {
      WEIGHT_T edge_weight = get_weight(graph.get_edge(current.id, neighbor));

      if (edge_weight < 0) {
        throw std::invalid_argument{
          ("Negative edge weight [" + std::to_string(edge_weight) + "] between vertices [" + std::to_string(current.id) + "] -> [" + std::to_string(neighbor) + "].")};
      }

      WEIGHT_T distance = current.dist_from_start + edge_weight;

      if (!vertex_info.contains(neighbor) ||
          distance < vertex_info[neighbor].dist_from_start) {
        vertex_info[neighbor] = {neighbor, distance, current.id};
        to_explore.push(vertex_info[neighbor]);
      }
```

@bobluppes  What do you think about it?

bobluppes:
Thanks for investigating @joweich and @MichaeINeumann. In that case let's move forward with the manual string concatenation. Do you want to do this in the current PR?

I will then create a ticket in our backlog to switch to `std::format` once we can reasonably assume that clients have access to compliant compilers.

Issue #53, opened 2023-08-06
[ALGO] Tarjan's Strongly Connected Components


bobluppes:
# Tarjan's Strongly Connected Components
Tarjan's algorithm can be used to find the [Strongly Connected Components](https://en.wikipedia.org/wiki/Strongly_connected_component) (SCCs) of a directed graph. An SCC is a subset of vertices in the graph for which every vertex is reachable from every other vertex in the subset. See the [wikipedia](https://en.wikipedia.org/wiki/Tarjan%27s_strongly_connected_components_algorithm) entry for more details on Tarjan's algorithm.

## Syntax
The algorithm should have the following syntax:

```c++
/**
 * Computes the Strongly Connected Components (SCCs) of a graph using Tarjan's algorithm.
 *
 * This function takes a graph and returns a vector of vectors representing the SCCs.
 * Each inner vector contains the vertices of a strongly connected component, and the outer vector
 * contains all the strongly connected components in the graph.
 *
 * @tparam V Vertex type.
 * @tparam E Edge type.
 * @param graph The graph for which to compute SCCs.
 * @return std::vector<std::vector<vertex_id_t>> A vector of vectors representing SCCs.
 */
template <typename V, typename E>
[[nodiscard]] std::vector<std::vector<vertex_id_t>> tarjans_strongly_connected_components(const graph<V, E, graph_type::DIRECTED>& graph);

```

This should live in the `graaf::algorithm` namespace under `include/graaflib/algorithm/strongly_connected_components.h`.

## Definition of Done
This issue is done when:

- [ ] The algorithm is implemented
- [ ] The new function has a javadoc-style comment explaining the interface
- [ ] Appropriate tests are added under `test/graaflib/algorithm/strongly_connected_components_test.cpp`
- [ ] A test coverage of at least 95% is reached
- [ ] A documentation entry is added under `docs/docs/algorithms` under the appropriate category
  - Just adding a short description and the algorithm syntax here is fine
  - See the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation) on how to build the documentation locally
- [ ] The algorithm is added to the list of algorithms in `README.md`

ndcroos:
Hi, I'd like to work on this after doing Bellman-Ford.

bobluppes:
Hi @ndcroos, awesome, I will assign it to you :)

PR #78, opened 2023-08-15
Work on Tarjan's SCC algorithm

ndcroos:
See https://github.com/bobluppes/graaf/issues/53

### DIFF
--- a/include/graaflib/algorithm/strongly_connected_components.tpp
+++ b/include/graaflib/algorithm/strongly_connected_components.tpp
@@ -1,0 +1,82 @@
+#pragma once
+
+#include <graaflib/algorithm/strongly_connected_components.h>
+
+#include <functional>  // For std::function
+#include <stack>
+#include <unordered_map>
+#include <vector>
+
+namespace graaf::algorithm {
+
+template <typename V, typename E>
+[[nodiscard]] std::vector<std::vector<vertex_id_t>>
+tarjans_strongly_connected_components(
+    const graph<V, E, graph_type::DIRECTED>& graph) {
+  // Vector to store strongly connected components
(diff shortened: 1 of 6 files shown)


include/graaflib/algorithm/strongly_connected_components.tpp
```diff
@@ -0,0 +1,82 @@
...
+  std::unordered_map<vertex_id_t, size_t> low_links;
+  std::unordered_map<vertex_id_t, bool> on_stack;
+
+  // Counter for indexing vertices
+  size_t index_counter = 0;
+
+  // Lambda function for the strong connect traversal
+  std::function<void(vertex_id_t)> strong_connect;
+
+  strong_connect = [&](vertex_id_t vertex) {
```
bobluppes, line 33: [very minor]: Since we are not passing around the lambda in an interface, we can keep it a lambda closure rather than assigning it to a function object:
```suggestion
  // Lambda function for the strong connect traversal
  const auto strong_connect = [&](vertex_id_t vertex) {
```

bobluppes (approved):
I will commit the minor suggestions, after which I will merge the PR. So no need to address them.

include/graaflib/algorithm/strongly_connected_components.tpp
bobluppes: Ah, I see now why you did this, since the function object is used in the body the compiler needs to know the type. Reverted this change in [35036c3](https://github.com/bobluppes/graaf/pull/78/commits/35036c3f6b39e9bca14dd62a9b97c30f350a11e2).

It might also be interesting to see if we can implement this in terms of the existing traversal algorithms in `graph_traversal.h`, but I will leave this for a follow up.

PR #80, opened 2023-08-17
Doc (dfs) add documentation for DFS (#57)

MichaeINeumann:
Hello, I am adding the documentation for DFS. #57 
It is quite the same as BFS, with a stack instead of a queue. 
Also the call is the same, except for the name, as far as I could see.

@bobluppes Sorry for the rebuild, I played a bit with git and did a rebase. somehow the old PR disappeared. #77


bobluppes (approved):
Thanks for working on this, looks great and very detailed 🚀 
LGTM

PR #86, opened 2023-09-01
Docs for BFS-based shortest path search

### LINKED ISSUE #58: [DOCS] Document BFS Based Shortest Path
# Documentation BFS Based Shortest Path
The goal of this issue if the improve the documentation of an existing algorithm. This documentation should go into the algorithms section of our [docusaurus docs](https://bobluppes.github.io/graaf/docs/algorithms/intro).

More detail on how to build the documentation locally can be found on the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation).

An existing documentation page exists, which should be extended. The page can be found under `docs/docs/algorithms/shortest-path/bfs-based-shortest-path.md`.

The corresponding algorithm is called `bfs_shortest_path` and can be found under `include/graaflib/algorithm/shortest_path.h`.

## Documentation Contents
The documentation entry should adhere to the following template:

```markdown
# [ALGORITHM_NAME]
- A short description of the algorithm, what does it do and in which use cases is it used.
- What are the limitations of the algorithm, does it work on both directed and undirected graphs? Does it consider edge weights, and if yes, does it support negative weights?
- Link to the wikipedia entry on the algorithm.

## Syntax
- A code block with the syntax of the algorithm. This should now include any javadoc comments.
- An explanation of the parameters (including template parameters) and the return type.

## (Optional) See also
- If there are similar algorithms, or we have a slightly different version of the same algorithm, you can link it here.
- If we have an example (found on the example page of the docs) which uses this algorithm, you can link it here.
```

joweich:
Closes #58

### DIFF
--- a/docs/docs/algorithms/shortest-path/bfs-based-shortest-path.md
+++ b/docs/docs/algorithms/shortest-path/bfs-based-shortest-path.md
@@ -3,3 +3,27 @@ sidebar_position: 1
 ---
 
 # BFS Based Shortest Path
+Breadth-First Search (BFS) is a graph traversal algorithm that efficiently finds the shortest
+path between two vertices in an **unweighted graph** by exploring vertices level by level,
+guaranteeing the shortest path, and has a time complexity of `O(|E| + |V|)`,
+where `|V|` is the number of vertices and `|E|` is the number of edges in the graph.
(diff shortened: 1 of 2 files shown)


bobluppes (approved):
Thanks for addressing this, looks great!

PR #93, opened 2023-09-05
Topological Sorting

Hromz:
Added TS DFS-based.
- Algorithm
- Tests
- Documentation
- Comments


docs/docs/algorithms/topological-sort/topological-sort.md
```diff
@@ -0,0 +1,21 @@
...
+The runtime of the algorithm is `O(|V|+|E|)` and the memory consumption is `O(|V|)`.
+
+[wikipedia](https://en.wikipedia.org/wiki/Topological_sorting)
+
+## Syntax
+
+```cpp
+template <typename V, typename E>
+[[nodiscard]] std::vector<edge_id_t> kruskal_minimum_spanning_tree(
+    const graph<V, E, graph_type::UNDIRECTED>& graph);
```
joweich, line 17:
```suggestion
template <typename V, typename E>
[[nodiscard]] std::vector<vertex_id_t> topological_sort(
    const graph<V, E, graph_type::DIRECTED>& graph);
```

docs/docs/algorithms/topological-sort/topological-sort.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+- **graph** The graph to traverse.
```
joweich, line 20:
```suggestion
- **graph** The directed graph to traverse.
```

include/graaflib/algorithm/topological_sorting.tpp
```diff
@@ -0,0 +1,60 @@
...
+        }
+    }
+
+        sorted_vertices.push_back(start_vertex);
+    }
+
+}; // namespace detail
+
+template <typename V, typename E>
+std::vector<vertex_id_t> topological_sort(
```
joweich, line 37: We should change the return type to use `std::optional<>` here. We can then return `std::nullopt` in case there are cycles.  See the shortest path algorithms where we do something very similar in case there is no shortest path.

include/graaflib/algorithm/topological_sorting.tpp
```diff
@@ -0,0 +1,60 @@
...
+    const graph<V, E, graph_type::DIRECTED>& graph) {
+    
+    // Graph should be acyclic
+    if (dfs_cycle_detection(graph)) {
+        return {};
```
joweich, line 42:
```suggestion
        return std::nullopt;    
```

joweich, line 42: Minor food for thought: Maybe there is a possibility to check for cycles during the topological sorting itself, which then might have better performance than the two-step approach. We can think about checking this in a follow up PR

include/graaflib/algorithm/topological_sorting.tpp
```diff
@@ -0,0 +1,60 @@
...
+    
+    // Graph should be acyclic
+    if (dfs_cycle_detection(graph)) {
+        return {};    
+    }
+
+    using enum detail::vertex_status;
+    std::vector<vertex_id_t> sorted_vertices{};
+    sorted_vertices.reserve(graph.vertex_count());
+    std::unordered_map<vertex_id_t, detail::vertex_status> processed_vertices{};
```
joweich, line 48: do we really _need_ the `vertex_status` enum? Wouldn't a container that holds the visited vertices suffice? Also, it looks like we never set (i.e. initialize) something to `UNVISITED`

include/graaflib/algorithm/topological_sorting.tpp
```diff
@@ -0,0 +1,60 @@
...
+    
+    processed_vertices[start_vertex] = vertex_status::VISITED;
+    for (const auto& next_vertex : graph.get_neighbors(start_vertex)) {
+        if (processed_vertices[next_vertex] == vertex_status::UNVISITED) {
+            do_dfs_topological_sort(graph, next_vertex, processed_vertices,
+                sorted_vertices);
+        }
+    }
+
+        sorted_vertices.push_back(start_vertex);
```
joweich, line 31:
```suggestion
    sorted_vertices.push_back(start_vertex);
```

joweich (commented):
Hey @Hromz, I did an initial round of review that primarily concerns codestyle improvements. Let me know what you think!

docs/docs/algorithms/topological-sort/topological-sort.md
Hromz: Oh, my bad, I missed this one.

docs/docs/algorithms/topological-sort/topological-sort.md
Hromz: Alright

include/graaflib/algorithm/topological_sorting.tpp
Hromz: Hmm sounds good ill work on it

include/graaflib/algorithm/topological_sorting.tpp
Hromz: Ok

Hromz: Agreed, in terms of O(n), it's still n, but in the case of a large number of vertices, it will affect performance.

include/graaflib/algorithm/topological_sorting.tpp
Hromz: Added for more readability. Let me check if the built-in container will do the job. But what about if the user doesn't want to fill the inner container?

include/graaflib/algorithm/topological_sorting.tpp
Hromz: Ok

Hromz:
Hey @joweich, thanks for your comments. I'll fix it.

include/graaflib/algorithm/topological_sorting.tpp
bobluppes: Good idea! I would indeed be fine with doing this in a follow-up, as the current PR already adds value 👍🏻

include/graaflib/algorithm/topological_sorting.tpp
bobluppes: Looking at how the container is used currently, the "initialization" is done by calling `operator[]`. Since this operator default constructs values for missing keys, they are initialized to the first enum value (`UNVISITED`).

Since this data structure is not exposed to the user of the algorithm, we can treat it as an implementation detail and freely change the underlying container type. So no need to provide multiple implementations.

What do you think about using an `std::unordered_set` here? When the vertex is visited you can insert into it and the rest of the code could do something like `processed_vertices.contains(current_vertex_id)`.

include/graaflib/algorithm/topological_sorting.tpp
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    using enum detail::vertex_status;
+    std::vector<vertex_id_t> sorted_vertices{};
+    sorted_vertices.reserve(graph.vertex_count());
+    std::unordered_map<vertex_id_t, detail::vertex_status> processed_vertices{};
+
+    for (const auto& vertex : graph.get_vertices()) {
+        if (processed_vertices[vertex.first] == UNVISITED) {
+            detail::do_dfs_topological_sort(graph, vertex.first, processed_vertices, sorted_vertices);
```
bobluppes, line 52: [very minor]: To make this slightly more readable, how about some structured binding here to unpack the pair:
```suggestion
    for (const auto& [vertex_id, _] : graph.get_vertices()) {
        if (processed_vertices[vertex.first] == UNVISITED) {
            detail::do_dfs_topological_sort(graph, vertex_id, processed_vertices, sorted_vertices);
```

test/graaflib/algorithm/topological_sorting_test.cpp
```diff
@@ -0,0 +1,192 @@
...
+  const auto vertex_3{graph.add_vertex(30)};
+  const auto vertex_4{graph.add_vertex(40)};
+
+  graph.add_edge(vertex_1, vertex_2, 25);
+  graph.add_edge(vertex_2, vertex_3, 35);
+  graph.add_edge(vertex_3, vertex_4, 45);
+
+  // WHEN;
+  auto sorted_vertices = topological_sort(graph);
+  std::vector<vertex_id_t> expected_vertices{0, 1, 2, 3};
```
bobluppes, line 35: [minor]: I would propose to use the obtained vertex IDs in the `expected_vertices`. Then your tests are completely decoupled from the logic which generates vertex IDs.
```suggestion
  const auto vertex_1{graph.add_vertex(10)};
  const auto vertex_2{graph.add_vertex(20)};
  const auto vertex_3{graph.add_vertex(30)};
  const auto vertex_4{graph.add_vertex(40)};

  graph.add_edge(vertex_1, vertex_2, 25);
  graph.add_edge(vertex_2, vertex_3, 35);
  graph.add_edge(vertex_3, vertex_4, 45);

  // WHEN;
  auto sorted_vertices = topological_sort(graph);
  std::vector<vertex_id_t> expected_vertices{vertex_1, vertex_2, vertex_3, vertex_4};
```

bobluppes: Same for the other tests

include/graaflib/algorithm/topological_sorting.tpp
bobluppes, line 31: If you add the `start_vertex` to the `sorted_vertices` before the `for`-loop, do you already have the vertices recorded in the correct order? Maybe it would be possible to get rid of the `std::reverse` then.

bobluppes (commented):
Looks very good already! Just some minor comments apart from @joweich's points :)

include/graaflib/algorithm/topological_sorting.tpp
Hromz: Oh, good advice. I guess I will correct my other PR. I wasn't satisfied with 'vertex.first' aswell!

include/graaflib/algorithm/topological_sorting.tpp
Hromz: I'm not sure about that. In this case, vertices will be stored as usual in DFS traversal, ignoring all other dependencies.
Consider this directed graph
       0
    /   \
   *     *
   1    2
   \    /
    *  *
      3
If we do preorder, the vertices will be 0–1–3–2, which is incorrect for TS.
But if we do post order 0 - 1 - 2 - 3, Or maybe I missed something, correct me if I'm getting something wrong.

include/graaflib/algorithm/topological_sorting.tpp
Hromz: I was thinking about how to combine cycle detection and node pushback traversal without using enums, etc. What do you think about using unordered_set as the return type from the function instead of vector? So basically, in the case of a cycle, vertex should be visited but wasn't present in the set for return type.

docs/docs/algorithms/topological-sort/topological-sort.md
```diff
@@ -0,0 +1,21 @@
...
+## Syntax
+
+```cpp
+template <typename V, typename E>
+[[nodiscard]] std::optional<std::vector<vertex_id_t>> topological_sort(
+    const graph<V, E, graph_type::DIRECTED>& graph);
+```
+
+- **graph** The directed graph to traverse.
+- **return** Vector of vertices sorted in topological order. If the graph contains cycles, it returns an empty vector.
```
joweich, line 21:
```suggestion
- **return** Vector of vertices sorted in topological order. If the graph contains cycles, it returns std::nullopt.
```

include/graaflib/algorithm/topological_sorting.tpp
joweich: I quickly checked @bobluppes's proposal and it indeed fails the tests. I would suggest going with the solution in this PR and think about ways getting rid of the `std::reverse` in a follow up.

joweich (commented):
@Hromz LGTM, all points in the reviews have been addressed properly :rocket: 
Only have one quick fix concerning the docs.

include/graaflib/algorithm/topological_sorting.tpp
bobluppes: Ah, my bad indeed. One thing we could investigate is to use a `std::deque` rather than a `vector` and call `push_front`. However, let's not optimize prematurely and let's tackle this in a follow up.

bobluppes (approved):
Also thanks to @joweich for the great feedback!

Issue #88, opened 2023-09-04
[ALGO] Graph Coloring (Greedy)


bobluppes:
# Greedy Graph Coloring
The Greedy Graph Coloring algorithm is a heuristic approach to assigning colors to the vertices of a graph in such a way that no adjacent vertices share the same color. It is a fundamental algorithm with applications in scheduling, map coloring, and various optimization problems. The goal is to implement this algorithm to add to the library's suite of graph algorithms.

See the [wikipedia entry](https://en.wikipedia.org/wiki/Greedy_coloring) for more details.

## Syntax
The algorithm should have the following syntax:

```c++
/**
 * @brief Greedy Graph Coloring Algorithm
 *
 * This function performs greedy graph coloring on a given graph. It assigns
 * colors to vertices in such a way that no two adjacent vertices share the
 * same color. The algorithm is heuristic and does not guarantee an optimal
 * coloring.
 *
 * @tparam GRAPH The type of the graph
 * @param graph The graph object
 * @return An unordered_map where keys are vertex identifiers and values are their respective colors.
 */
template <typename GRAPH>
std::unordered_map<vertex_id_t, int> greedy_graph_coloring(const GRAPH& graph);

```

This should live in the `graaf::algorithm` namespace under `include/graaflib/algorithm/coloring/greedy_graph_coloring.h`.

## Definition of Done
This issue is done when:

- [ ] The algorithm is implemented
- [ ] The new function has a javadoc-style comment explaining the interface
- [ ] Appropriate tests are added under `test/graaflib/algorithm/coloring/greedy_graph_coloring_test.cpp`
- [ ] A test coverage of at least 95% is reached
- [ ] A documentation entry is added under `docs/docs/algorithms` under the appropriate category. The category should follow the file path of the new algorithm.
  - In case the category does not exist yet, please add one
  - Just adding a short description and the algorithm syntax here is fine
  - See the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation) on how to build the documentation locally
- [ ] The algorithm is added to the list of algorithms in `README.md`

ndcroos:
Hi, I'd like to do this. Can you assign it to me?

bobluppes:
> Hi, I'd like to do this. Can you assign it to me?

Hi, of course, would be happy to :)

PR #94, opened 2023-09-11
Work on greedy graph coloring.

ndcroos:
See https://github.com/bobluppes/graaf/issues/88.


ndcroos:
I removed the empty dir 'coloring' and renamed 'greedy_graph_coloring_test.cpp' to 'graph_coloring_test.cpp'
If a separate dir is used that contains the tests, it seems that it won't execute without other changes.

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
```diff
@@ -0,0 +1,45 @@
...
+
+  // Get the vertices from the graph
+  const auto& vertices = graph.get_vertices();
+
+  // Iterate through each vertex
+  for (const auto& vertex_pair : vertices) {
+    const vertex_id_t& current_vertex_id = vertex_pair.first;
+
+    // Create a set of colors used by neighboring vertices
+    std::unordered_set<int> neighbor_colors;
```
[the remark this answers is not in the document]
bobluppes: An example situation where this would matter:

```c++
struct foo {
  int i; // built-in type
  std::string s;  // class type
}

foo bar;  // will initialize s, but will not initialize i
foo baz{}; // initializes both s and i
```

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
```diff
@@ -0,0 +1,45 @@
...
+std::unordered_map<vertex_id_t, int> greedy_graph_coloring(const GRAPH& graph) {
+  // Initialize a map to store the coloring
+  std::unordered_map<vertex_id_t, int> coloring;
+
+  // Get the vertices from the graph
+  const auto& vertices = graph.get_vertices();
+
+  // Iterate through each vertex
+  for (const auto& vertex_pair : vertices) {
+    const vertex_id_t& current_vertex_id = vertex_pair.first;
```
bobluppes, line 19: [minor]: Another way of doing this would be to use structured binding to unpack the pair:

```suggestion
  for (const auto& [current_vertex_id, _] : vertices) {
```

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
```diff
@@ -0,0 +1,45 @@
...
+
+    // Create a set of colors used by neighboring vertices
+    std::unordered_set<int> neighbor_colors;
+
+    // Iterate through neighboring vertices
+    for (const auto& neighbor_id : graph.get_neighbors(current_vertex_id)) {
```
bobluppes, line 25: [very minor]: As the `vertex_id_t` is trivially copyable (it is a `size_t`), we can take it by value rather than by reference to save one layer of indirection:
```suggestion
    // Iterate through neighboring vertices
    for (const auto neighbor_id : graph.get_neighbors(current_vertex_id)) {
```

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
```diff
@@ -0,0 +1,45 @@
...
+      auto neighbor_color_it = coloring.find(neighbor_id);
+      if (neighbor_color_it != coloring.end()) {
+        neighbor_colors.insert(neighbor_color_it->second);
+      }
+    }
+
+    // Find the smallest available color for the current vertex
+    int available_color = 1;
```
bobluppes, line 33: [very minor]: Mostly a definition question, but what do you think about starting the coloring at 0?

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
```diff
@@ -0,0 +1,45 @@
...
+    while (neighbor_colors.count(available_color) > 0) {
+      ++available_color;
+    }
```
bobluppes, line 36: Since we do not seem to really need the set here, what do you think about something like:

```c++
int available_color{0};
for (const auto neighbor_id : graph.get_neighbors(current_vertex_id)) {
  if (coloring.contains(neighbor_id)) {
    const auto neighbor_color{coloring.at(neighbor_id)};
    if (neighbor_color >= available_color) {
      available_color = neighbor_color + 1;
    }
  }
}
```

docs/docs/algorithms/coloring/greedy-graph-coloring.md
```diff
@@ -0,0 +1,29 @@
...
+
+If the graph has different connected components, each component will be treated as a separate simple connected graph.
+
+The algorithm is heuristic and does not guarantee an optimal number of different colors (that is, equal to the chromatic number of a simple, connected graph).
+
+Colors are represented by the numbers 0, 1, 2,... The greedy algorithm considers the vertices of the graph in sequence and assigns each vertex its first available color, i.e. the color with the smallest number that is not already used by one of its neighbors.
+
+The overall worst-case time complexity of the algorithm is `O(n^2)`. In cases where the graph has a fixed degree (a constant number of neighbors for each vertex), the time complexity can be approximated as `O(n)`. However, if the graph is highly connected (dense) and approaches a complete graph, the time complexity could approach `O(n^2)`.
+
+If no coloring is possible, an empty `unordered_map` is returned.
```
bobluppes, line 17: Would it ever be possible that we find no coloring?

bobluppes (commented):
Thanks for working on this, looks very good already!
Mostly some style related feedback

bobluppes:
> I removed the empty dir 'coloring' and renamed 'greedy_graph_coloring_test.cpp' to 'graph_coloring_test.cpp' If a separate dir (coloring) is used for contain different files with test cases (to reflect the directory structure in the include dir), it seems that it won't execute the tests in these files without other changes.

Thanks for reporting! Sounds like I will have to adjust the CMake setup in order for this to work. I will do so in a follow-up

docs/docs/algorithms/coloring/greedy-graph-coloring.md
ndcroos: I think that the only case where we can't find a coloring, is when the graph variable doesn't have vertices.
So, I think in the case of an empty graph, an empoty unordered map should be returned.

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
ndcroos: Starting coloring at 0 is a good idea, one reason is that starting indexing at 0 is also used for vertex_id_t.

bobluppes:
closes #88

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
```diff
@@ -0,0 +1,41 @@
...
+  // Initialize a map to store the coloring
+  std::unordered_map<vertex_id_t, int> coloring{};
+
+  // Get the vertices from the graph
+  const auto& vertices = graph.get_vertices();
+
+  // Iterate through each vertex
+  for (const auto& [current_vertex_id, _] : vertices) {
+    // Create a set of colors used by neighboring vertices
+    std::unordered_set<int> neighbor_colors{};
```
bobluppes, line 20: [very minor]: I think we can now get rid of this, as it is unused:
```suggestion
```

docs/docs/algorithms/coloring/greedy-graph-coloring.md
bobluppes: Thanks, that makes sense!

bobluppes (approved):
Thanks for addressing the feedback, LGTM 🎉

Issue #60, opened 2023-08-06
[DOCS] Document DFS Based Cycle Detection


bobluppes:
# Documentation DFS Based Cycle Detection
The goal of this issue if the improve the documentation of an existing algorithm. This documentation should go into the algorithms section of our [docusaurus docs](https://bobluppes.github.io/graaf/docs/algorithms/intro).

More detail on how to build the documentation locally can be found on the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation).

An existing documentation page exists, which should be extended. The page can be found under `docs/docs/algorithms/cycle-detection/dfs-based.md`.

The corresponding algorithm is called `dfs_cycle_detection` and can be found under `include/graaflib/algorithm/cycle_detection.h`. There are two versions of the algorithm, one for directed graphs and one for undirected graphs. Both should be documented on the same page, where we list both available syntaxes.

## Documentation Contents
The documentation entry should adhere to the following template:

```markdown
# [ALGORITHM_NAME]
- A short description of the algorithm, what does it do and in which use cases is it used.
- What are the limitations of the algorithm, does it work on both directed and undirected graphs? Does it consider edge weights, and if yes, does it support negative weights?
- Link to the wikipedia entry on the algorithm.

## Syntax
- A code block with the syntax of the algorithm. This should now include any javadoc comments.
- An explanation of the parameters (including template parameters) and the return type.

## (Optional) See also
- If there are similar algorithms, or we have a slightly different version of the same algorithm, you can link it here.
- If we have an example (found on the example page of the docs) which uses this algorithm, you can link it here.
```

Hromz:
Hey @bobluppes, assign to me please. I wrote it so docs from me is good way of contributing, I guess.

bobluppes:
> Hey @bobluppes, assign to me please. I wrote it so docs from me is good way of contributing, I guess.

Hi, of course, much appreciated! Assigning this to you :)

Hromz:
Hey @bobluppes, check #96

PR #96, opened 2023-09-14
DOCS DFS cycle detection

Hromz:
- Short description
- Potential uses
- Limitation
- Syntax
- Optional


bobluppes:
closes #60

docs/docs/algorithms/cycle-detection/dfs-based.md
```diff
@@ -2,4 +2,55 @@
 sidebar_position: 1
 ---
 
-# DFS Based Cycle Detection
\ No newline at end of file
+# DFS Based Cycle Detection
+A DFS based cycle detection algorithm is used to identify cycles in graphs, both directed and undirected.
+The algorithm doesn't work on edges with weight and therefore cannot be used for negative weight cycles.
+
```
bobluppes, line 8: [minor]: It sounds like this is a disadvantage of the algorithm, but I am wondering if negative weight cycles would ever influence cycle detection. I would argue the algorithm does work for weighted graphs, but the edge weight will simply not be considered 🙈 

What do you think about this?

```suggestion
The algorithm can be used to detect cycles in the structure of a graph, as it does not consider edge weights.

```

bobluppes (approved):
Looks very good!
Only one question on which I would value your opinion. Looks good to merge otherwise 👍🏻

docs/docs/algorithms/cycle-detection/dfs-based.md
Hromz: I agree that your suggestion is more correct and a bit more readable.

bobluppes (approved):
Thanks, LGTM

PR #98, opened 2023-09-24
Floyd-Warshall algorithm

Hromz:
Added Floyd-Warshall alogorithm, tests and docs.


include/graaflib/algorithm/shortest_path/floyd_warshall.tpp
```diff
@@ -0,0 +1,77 @@
...
+
+};  // namespace detail
+
+template <typename V, typename E, graph_type T, typename WEIGHT_T>
+std::vector<std::vector<WEIGHT_T>> floyd_warshall_shortest_paths(
+    const graph<V, E, T>& graph) {
+  WEIGHT_T ZERO{};
+  std::size_t n = graph.vertex_count();
+  auto INF = std::numeric_limits<WEIGHT_T>::max();
+  auto NO_PATH = std::numeric_limits<WEIGHT_T>::min();
```
bobluppes, line 39: Regarding using `std::numeric_limits<WEIGHT_T>::min()` for vertex pairs between which we cannot find a path (i.e. vertices in negative weight cycles), consider the case where the `WEIGHT_T` is an `unsigned` type. 

In that case, the caller cannot distinguish between a `NO_PATH` and a true distance of `0`. However, I do not see an elegant way of handling this. What do you think about removing the negative cycle detection, as it is also documented on the interface that the algorithm does not work in this case.

bobluppes (commented):
Looks very good already! 
I will take a closer look tomorrow but wanted to provide some early feedback :)

include/graaflib/algorithm/shortest_path/floyd_warshall.tpp
Hromz: Oh, I didn't see this one. My bad. I'll fix it right away. Agreed, we could move it to a different PR as Floyd-Warshall negative cycle detection so results will be less confusing. And in this case, NO_PATH will no longer be needed.

I look forward to your next comments about PR. :)

bobluppes (approved):
Thanks for working on this, LGTM 🚀

PR #99, opened 2023-09-24
[TOOL] Add Dependabot for GitHub actions

joweich:
I found this to be quite handy to keep gh actions up-to-date. We could think of extending it with checking cpp specific dependencies down the road as well.


bobluppes (approved):
LGTM!

Have never used this before, but it looks pretty cool!

PR #119, opened 2023-10-01
Implementation of Koraraju's Algorithm

toadkarter:
* [ ]  The algorithm is implemented
* [x]  The new function has a javadoc-style comment explaining the interface
* [ ]  Appropriate tests are added under `test/graaflib/algorithm/strongly_connected_components/kosarajus_test.cpp`
* [ ]  A test coverage of at least 95% is reached
* [ ]  A documentation entry is added under `docs/docs/algorithms` under the appropriate category 
* [ ]  The algorithm is added to the list of algorithms in `README.md`

### DIFF
--- a/include/graaflib/algorithm/strongly_connected_components/kosarajus.h
+++ b/include/graaflib/algorithm/strongly_connected_components/kosarajus.h
@@ -1,0 +1,28 @@
+#pragma once
+
+#include <graaflib/types.h>
+#include <graaflib/graph.h>
+
+namespace graaf::algorithm {
+
+/**
+ * @brief Finds strongly connected components in a directed graph using Kosaraju's algorithm.
+ *
+ * Kosaraju's algorithm identifies strongly connected components (SCCs) in a directed graph.
+ * The algorithm uses two depth-first searches to discover these components.
+ *
+ * @tparam V The vertex type of the graph.
+ * @tparam E The edge type of the graph.
+ * @param graph The input directed graph.
+ *
+ * @return A std::vector of vectors, each of which contains the vertex IDs forming
+ *         a strongly connected component.
+ */
+template <typename V, typename E>
+std::vector<std::vector<vertex_id_t>> kosarajus_strongly_connected_components(
+    const graph<V, E, graph_type::DIRECTED>& graph
+);
+
+} // namespace graaf::algorithm
+
+#include "kosarajus.tpp"
--- a/include/graaflib/algorithm/strongly_connected_components/kosarajus.tpp
+++ b/include/graaflib/algorithm/strongly_connected_components/kosarajus.tpp
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <graaflib/algorithm/graph_traversal/depth_first_search.h>
+#include <graaflib/algorithm/strongly_connected_components/kosarajus.h>
+
+#include <stack>
+
+namespace graaf::algorithm {
+
+template <typename V, typename E>
+std::vector<std::vector<vertex_id_t>> kosarajus_strongly_connected_components(
+    const graph<V, E, graph_type::DIRECTED>& graph) {
+  // Vector to store strongly connected components
+  std::vector<std::vector<vertex_id_t>> sccs;
+
+  // Stack to hold vertices during traversal
+  std::stack<vertex_id_t> stack;
+}
+
+}  // namespace graaf::algorithm
--- a/include/graaflib/graph.h
+++ b/include/graaflib/graph.h
@@ -218,6 +218,11 @@ using directed_graph = graph<VERTEX_T, EDGE_T, graph_type::DIRECTED>;
 template <typename VERTEX_T, typename EDGE_T>
 using undirected_graph = graph<VERTEX_T, EDGE_T, graph_type::UNDIRECTED>;
 
+template <typename VERTEX_T, typename EDGE_T>
+void get_transposed_graph(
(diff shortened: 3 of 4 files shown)


toadkarter:
@bobluppes  Hi there, I'm just working my way through the transpose function and wanted to see if you had any ideas as to the best way of testing this function. 

I have set this function up currently as a free-standing function, as you mentioned, and it uses a similar signature to that of the boost equivalent - you pass in an in parameter, your existing graph, and an out parameter, which is the transposed graph. 

Now, the issue that I have is that obviously the ids of the vertices are all assigned internally, and there's no way of, as far as I am aware, checking to see if there is an edge with particular values - only vertex ids. This makes assertions while writing tests quite difficult, so just wanted to see if you had any thoughts on how to approach this. 

Also please do let me know if the way I have approached this so far suits or if you would like me to make any changes - this is currently still a draft of course but if I am going off course do let me know.

toadkarter:
Looking at this again, and I'm not sure if it's going to be possible to have a `get_transposed_graph` function to exist outside the graph class. As far as I can see, it is impossible for us to create a new graph (i.e. the transposed one) that is going to have the exact same vertex and edge ids as another graph, as the vertex / edge id assignments are all handled internally.

What I propose we do instead is create a new function on the class called `transpose()` which just looks through the `edges_` map and flips the vertices inside it. This seems like the cleanest way of doing this without additional complexity. Please do let me know what you think!

bobluppes:
> Looking at this again, and I'm not sure if it's going to be possible to have a `get_transposed_graph` function to exist outside the graph class. As far as I can see, it is impossible for us to create a new graph (i.e. the transposed one) that is going to have the exact same vertex and edge ids as another graph, as the vertex / edge id assignments are all handled internally.
> 
> What I propose we do instead is create a new function on the class called `transpose()` which just looks through the `edges_` map and flips the vertices inside it. This seems like the cleanest way of doing this without additional complexity. Please do let me know what you think!

Hi @toadkarter, thanks for working on this! I see what you mean indeed. What do you think about extending the public interface of `graph` such that we have setters for vertices where we also manually supply the edge ID?

toadkarter:
@bobluppes Interesting, so would you be suggesting something along the lines of `set_edge_vertex_ids(edge_id, vertex_id_lhs, vertex_id_rhs)` (roughly speaking)? If you don't mind exposing that kind of functionality to the library user then I think that will work! I think it should be ok to expose this though, given that if the user uses this function, they will be aware that they are mutating the graph structure.

bobluppes:
> @bobluppes Interesting, so would you be suggesting something along the lines of `set_edge_vertex_ids(edge_id, vertex_id_lhs, vertex_id_rhs)` (roughly speaking)? If you don't mind exposing that kind of functionality to the library user then I think that will work! I think it should be ok to expose this though, given that if the user uses this function, they will be aware that they are mutating the graph structure.

Indeed! And instead of altering the existing edge IDs, what do you think about specifying the vertex ID already at creation?
```c++
vertex_id_t add_vertex(auto&& vertex, vertex_id_t id);
```

toadkarter:
> 
> Indeed! And instead of altering the existing edge IDs, what do you think about specifying the vertex ID already at creation?
> 
> ```c++
> vertex_id_t add_vertex(auto&& vertex, vertex_id_t id);
> ```

Hmm, so this will be an overload to the `add_vertex` function already in `graph` class, with the caveat that we are now specifying the exact id that we want to map the relevant value to? If so, then I guess we would probably want to throw an exception if the supplied ID is not currently in the graph. We could run into issues where a user does not handle such an exception but if we make this clear in the interface / documentation then I think it should be ok! And then the transpose function should flow quite naturally from this once this overloaded `add_vertex` is implemented. 

Just seeing that there are going to be quite a few changes in this PR, would you like me to open a new one for the work on `transpose` or are you happy to me just to keep it all together with the Kosaraju's Algorithm? I don't mind either way, whatever works best for you!

bobluppes:
> And

Indeed, I think it makes sense to throw an exception if the input is invalid. However, shouldn't we throw if the supplied ID **is** currently in the graph?

Adding these new features in separate PRs sounds like a good idea and works for me 👍🏻

toadkarter:
> Indeed, I think it makes sense to throw an exception if the input is invalid. However, shouldn't we throw if the supplied ID **is** currently in the graph?

Ah, that's my bad - I had initially thought that the goal of `add_vertex` was to override the existing vertex id with a new value. I think if that's not the case, then wouldn't we potentially have issues with colliding IDs? Say for example that I have a graph with vertex IDs 0, 1, 2, which I have added using the regular method. The ids have been provided to me with `vertex_id_supplier_`, which is now set to "2". I then use the new method to add a vertex with ID 3. Next, I continue adding vertices to my graph with the _old version_ of `add_vertex`. At this point, `vertex_id_supplier_` will increment and give me ID "3", which now causes a collision because we already have a vertex with this ID (which we have added manually) in the graph.   

What we could do to avoid this, I suppose, is when we try to retrieve a new ID using `vertex_id_supplier_` we can just check to see if such an id is not already in the vertex map, and if it is, we continue to increment `vertex_id_supplier_` until we get an id that isn't taken? The likelihood of this happening are slim I'm sure but it is an edge case that I think we should take into account. If this works then I will also write a test for this scenario!  

> Adding these new features in separate PRs sounds like a good idea and works for me 👍🏻

Sounds great, will work on this in a separate PR so.

toadkarter:
I will get cracking on this one now, but will open up a new PR just to keep things clean!

PR #122, opened 2023-10-01
Updated readme.md

LinhNguyenLe2109:
### Changes
- Moved About section to a new section call "How to use Graaf". I believe this title suit that section more
- Add a new About section. When I first read the project I had no idea what this one was about. I had to search online for a definition
- Moved "Installation" Section above "How to use...". I think this fits with the flow of most project more.
- Updated all links in the "Algorithms section". Some of them are wrong urls, I updated them and added other links to improve user experince
- Add a contributors list. It would be nice to see a list of people working on the project. Let me know if you think it's unnecessary

### New found issues:
- I find the the installation guide is not consistent, the one on README and the one on the doc are quite different. So what should I use? Should it be on the doc or the one on README? I believe we need clearer instruction.
- In "How to use Graaf section", should we have a clearer guide and let the user actually create and see the result of the project? It would be nice if the user can perform some simple algorithms with their newly created graph

Let me know if you need any change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,27 @@ is designed as a lightweight alternative to the [Boost Graph Library (BGL)](http
 Graaf is written in modern C++20 and is designed with generality in mind. As such, it can be used to store any
 user-defined vertex and edge class as well as allow users to perform complex algorithms.
 
+# Installation
+
+The most straightforward way to use the Graaf in your project is to include it as a header-only library. Please take a
+look at the [installation guide](https://bobluppes.github.io/graaf/docs/quickstart/installation) for alternative
+installation methods.
+
+## Header-Only Library
+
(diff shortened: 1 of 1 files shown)


LinhNguyenLe2109:
@bobluppes HI can you see if the changes I made sufficent?

bobluppes:
> I find the the installation guide is not consistent, the one on README and the one on the doc are quite different. So what should I use? Should it be on the doc or the one on README? I believe we need clearer instruction.
    In "How to use Graaf section", should we have a clearer guide and let the user actually create and see the result of the project? It would be nice if the user can perform some simple algorithms with their newly created graph

- Good point on the installation guide, we should definitely update that 😅 I created #125 for this
- Regarding the "How to use Graaf section", that is also a good idea. Maybe we can reuse one of the [examples](https://bobluppes.github.io/graaf/docs/category/basic-examples) there?

PR #124, opened 2023-10-03
[TEST] Extract common test fixtures

### LINKED ISSUE #111: Extract common test fixtures
We use the [GoogleTest](http://google.github.io/googletest/) framework for our unit tests. Since the Graaf library supports arbitrary vertex and edge types, we make use of `TYPED_TEST_SUITE`s to run our tests for different types of graph instances (i.e. `directed_graph<int, int>`, `directed_graph<long, std::string>`, ...).

Currently, most tests redefine the same fixtures over and over. For instance, consider `dijkstra_shortest_path_test.cpp` and `a_star_test.cpp`, which both define two sets of types (`weighted_graph_types` and `weighted_graph_signed_types`) in their anonymous namespace.

The goal of this issue is to extract these `types` and `fixture` definitions to a separate file `test/utils/fixtures.h`. This should make it easier to write tests in the future, as we can simply reuse the existing fixtures.

I would give all test files a once over to identify the fixtures we can extract, but in general I think all fixtures can be consolidated into either a fixture containing unsigned edge weights or a fixture containing signed edge weights.

More info on `gtest`'s TYPED_TEST_SUITE can be found [here](http://google.github.io/googletest/reference/testing.html#TYPED_TEST_SUITE).

joweich:
Closes #111 
Just refactoring - no logic changes.

### DIFF
--- a/test/graaflib/algorithm/coloring/greedy_coloring_test.cpp
+++ b/test/graaflib/algorithm/coloring/greedy_coloring_test.cpp
@@ -2,6 +2,7 @@
 #include <graaflib/graph.h>
 #include <graaflib/types.h>
 #include <gtest/gtest.h>
+#include <utils/fixtures/fixtures.h>
 
 #include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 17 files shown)


bobluppes (approved):
Very nice, LGTM!

Also nice to get rid of so many unused includes. I think they must have snuck in during the latest refactoring 🙈

PR #136, opened 2023-10-03
Adding new utils header containing "get_transposed_graph" function

toadkarter:
**CHANGES**
* New overload for `add_vertex()` in the `graph` class has been added that allows user to specify an ID for a given vertex. 
* `get_transposed_graph()` function has been added, allowing the user to get a transposed version of a particular directed graph. 
* Both functions have a javadoc style comment explaining the interface.
* Tests have been added to both functions.

Note: There does not seem to be a section in the documentation explaining the `graph` interface generally outside of the Quickstart guide, so I have not made any changes to the documentation. However, if something like this is required I would be happy to make additions. 

If you would like me to make any changes please do let me know.

### DIFF
--- a/include/graaflib/algorithm/utils.h
+++ b/include/graaflib/algorithm/utils.h
@@ -8,8 +8,7 @@ namespace graaf {
  * Get transposed version of a given directed graph
  *
  * @param  graph The directed graph that is to be transposed
- * @param  out_transposed_graph An out parameter that will contain the
- * transposed graph
+ * @return directed_graph<VERTEX_T, EDGE_T> The transposed graph
  */
 template <typename VERTEX_T, typename EDGE_T>
 directed_graph<VERTEX_T, EDGE_T> get_transposed_graph(
--- a/include/graaflib/algorithm/utils.h
+++ b/include/graaflib/algorithm/utils.h
@@ -1,0 +1,20 @@
+#pragma once
+
+#include <graaflib/graph.h>
+
+namespace graaf {
+
+/**
+ * Get transposed version of a given directed graph
+ *
+ * @param  graph The directed graph that is to be transposed
+ * @param  out_transposed_graph An out parameter that will contain the
+ * transposed graph
+ */
+template <typename VERTEX_T, typename EDGE_T>
+directed_graph<VERTEX_T, EDGE_T> get_transposed_graph(
+    const directed_graph<VERTEX_T, EDGE_T>& graph);
+
+}  // namespace graaf
+
+#include "utils.tpp"
--- a/include/graaflib/graph.h
+++ b/include/graaflib/graph.h
@@ -180,6 +180,16 @@ class graph {
   [[nodiscard]] vertex_id_t add_vertex(auto&& vertex);
 
   /**
+   * Add a vertex to the graph with a specific ID
+   *
+   * @param  vertex The vertex to be added
+   * @param  id The requested ID for the new vertex
+   * @return vertices_id_t - The ID of the new vertex
+   * @throws id_taken exception - If the relevant ID is already in use
+   */
+  vertex_id_t add_vertex(auto&& vertex, vertex_id_t id);
+
+  /**
    * Remove a vertex from the graph and update all its neighbors
    *
    * @param  vertex_id - The ID of the vertex
@@ -217,7 +227,6 @@ using directed_graph = graph<VERTEX_T, EDGE_T, graph_type::DIRECTED>;
 
 template <typename VERTEX_T, typename EDGE_T>
 using undirected_graph = graph<VERTEX_T, EDGE_T, graph_type::UNDIRECTED>;
-
 }  // namespace graaf
 
 #include "graph.tpp"
--- a/include/graaflib/graph.tpp
+++ b/include/graaflib/graph.tpp
@@ -129,10 +129,24 @@ graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::get_neighbors(
 
 template <typename VERTEX_T, typename EDGE_T, graph_type GRAPH_TYPE_V>
 vertex_id_t graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::add_vertex(auto&& vertex) {
-  // TODO: check overflow
-  const auto vertex_id{vertex_id_supplier_++};
+  while (has_vertex(vertex_id_supplier_)) {
+    ++vertex_id_supplier_;
+  }
+  const auto vertex_id{vertex_id_supplier_};
   vertices_.emplace(vertex_id, std::forward<VERTEX_T>(vertex));
   return vertex_id;
+}
+
+template <typename VERTEX_T, typename EDGE_T, graph_type GRAPH_TYPE_V>
+vertex_id_t graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::add_vertex(auto&& vertex,
+                                                              vertex_id_t id) {
+  if (has_vertex(id)) {
+    throw std::invalid_argument{"Vertex already exists at ID [" +
+                                std::to_string(id) + "]"};
+  }
+
+  vertices_.emplace(id, std::forward<VERTEX_T>(vertex));
+  return id;
 }
 
 template <typename VERTEX_T, typename EDGE_T, graph_type GRAPH_TYPE_V>
--- a/test/graaflib/graph_test.cpp
+++ b/test/graaflib/graph_test.cpp
@@ -38,6 +38,13 @@ TYPED_TEST(GraphTest, VertexCount) {
   ASSERT_EQ(graph.vertex_count(), 2);
   ASSERT_TRUE(graph.has_vertex(vertex_id_2));
   ASSERT_EQ(graph.get_vertex(vertex_id_2), 20);
+
+  // WHEN - THEN
+  constexpr int specific_id = 2;
+  const auto vertex_specific_id{graph.add_vertex(30, specific_id)};
+  ASSERT_EQ(graph.vertex_count(), 3);
+  ASSERT_TRUE(graph.has_vertex(specific_id));
+  ASSERT_EQ(graph.get_vertex(specific_id), 30);
 }
 
 TYPED_TEST(GraphTest, RemoveVertex) {
@@ -173,6 +180,24 @@ TYPED_TEST(GraphTest, VertexTests) {
   EXPECT_FALSE(graph.has_vertex(nonExistingVertexId));
   EXPECT_TRUE(graph.has_vertex(vertex_id_2));
   EXPECT_EQ(graph.get_vertex(vertex_id_2), 20);
+
+  ASSERT_THROW(
+      {
+        try {
+          // Add vertex to ID that already exists
+          [[maybe_unused]] const auto duplicate_id{
+              graph.add_vertex(50, vertex_id_1)};
+          FAIL()
+              << "Expected std::invalid_argument exception, but no exception "
+                 "was thrown.";
+        } catch (const std::invalid_argument &ex) {
+          EXPECT_EQ(ex.what(), fmt::format("Vertex already exists at ID [{}]",
+                                           vertex_id_1));
+          throw;
+        }
+      },
+      std::invalid_argument);
+  EXPECT_EQ(graph.get_vertex(vertex_id_1), 1);
 }
 
 TYPED_TEST(GraphTest, GetEdgeNonExistingEdge) {
--- a/include/graaflib/algorithm/utils.tpp
+++ b/include/graaflib/algorithm/utils.tpp
@@ -1,0 +1,32 @@
+#pragma once
+
+namespace graaf {
+
+template <typename VERTEX_T, typename EDGE_T>
+directed_graph<VERTEX_T, EDGE_T> get_transposed_graph(
+    const directed_graph<VERTEX_T, EDGE_T>& graph) {
+  directed_graph<VERTEX_T, EDGE_T> transposed_graph{};
+  for (auto [edge_vertices, edge_type] : graph.get_edges()) {
+    const auto [vertex_id_lhs, vertex_id_rhs] = edge_vertices;
+
+    auto vertex_value_lhs = graph.get_vertex(vertex_id_lhs);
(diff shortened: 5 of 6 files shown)


include/graaflib/graph.h
```diff
@@ -218,6 +228,18 @@ using directed_graph = graph<VERTEX_T, EDGE_T, graph_type::DIRECTED>;
...
+ * Get transposed version of a given directed graph
+ *
+ * @param  graph The directed graph that is to be transposed
+ * @param  out_transposed_graph An out parameter that will contain the
+ * transposed graph
+ */
+template <typename VERTEX_T, typename EDGE_T>
+void get_transposed_graph(
+    const directed_graph<VERTEX_T, EDGE_T>& graph,
+    directed_graph<VERTEX_T, EDGE_T>& out_transposed_graph);
```
bobluppes, line 241: What do you think about moving this to a separate source file `include/graaflib/algorithm/utils.h`?

test/graaflib/graph_test.cpp
```diff
@@ -222,4 +255,28 @@ TYPED_TEST(GraphTest, ConstGetter) {
   ASSERT_NO_THROW(test_getters_on_const_graph(graph));
 }
 
+TEST(GraphTest, Transpose) {
+  // GIVEN
+  using graph_t = directed_graph<int, int>;
+  graph_t graph{};
+  const auto vertex_id_1 = graph.add_vertex(1);
```
bobluppes, line 262: Same for this test, we can then also move it to its own file

include/graaflib/graph.h
bobluppes, line 241: Instead of having `out_transposed_graph` an output parameter, I would rather use a return value:

```c++
directed_graph<V, E> get_transposed_graph(const directed_graph<V, E>& graph) {...}
```

include/graaflib/graph.tpp
```diff
@@ -201,4 +215,29 @@ void graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::remove_edge(
...
+    const directed_graph<VERTEX_T, EDGE_T>& graph,
+    directed_graph<VERTEX_T, EDGE_T>& out_transposed_graph) {
+  for (const auto [edge_vertices, edge] : graph.get_edges()) {
+    const auto [vertex_id_lhs, vertex_id_rhs] = edge_vertices;
+
+    auto vertex_value_lhs = graph.get_vertex(vertex_id_lhs);
+    auto vertex_value_rhs = graph.get_vertex(vertex_id_rhs);
+    auto edge_type = edge;
+
+    if (!out_transposed_graph.has_vertex(vertex_id_lhs)) {
```
bobluppes, line 229: What is the point of these checks? If the out graph already contains these vertices, shouldn't we throw an exception? Or better, could we construct a new empty graph in the function body and use that as the return value?

Then we can ensure ourselves that a vertex with that ID is not there yet

include/graaflib/graph.tpp
```diff
@@ -201,4 +215,29 @@ void graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::remove_edge(
...
+      auto vertex_id =
+          out_transposed_graph.add_vertex(vertex_value_lhs, vertex_id_lhs);
+    }
+
+    if (!out_transposed_graph.has_vertex(vertex_id_rhs)) {
+      auto vertex_id =
+          out_transposed_graph.add_vertex(vertex_value_rhs, vertex_id_rhs);
+    }
```
bobluppes, line 237: Since the vertex value could be expensive to copy, we could use moves here 😉 

```suggestion
    if (!out_transposed_graph.has_vertex(vertex_id_lhs)) {
      auto vertex_id =
          out_transposed_graph.add_vertex(std::move(vertex_value_lhs), vertex_id_lhs);
    }

    if (!out_transposed_graph.has_vertex(vertex_id_rhs)) {
      auto vertex_id =
          out_transposed_graph.add_vertex(std::move(vertex_value_rhs), vertex_id_rhs);
    }
```

include/graaflib/graph.h
toadkarter: > What do you think about moving this to a separate source file `include/graaflib/algorithm/utils.h`?

Brill, yeah I think that makes sense. That way we are not polluting the graph headers with auxiliary files that the users might not not need.

test/graaflib/graph_test.cpp
toadkarter: No prob, will do!

include/graaflib/graph.h
toadkarter: No problem! Will make this tweak

include/graaflib/graph.tpp
toadkarter: > What is the point of these checks? If the out graph already contains these vertices, shouldn't we throw an exception? Or better, could we construct a new empty graph in the function body and use that as the return value?
> 
> Then we can ensure ourselves that a vertex with that ID is not there yet

I think what I was trying to get at here was a situation like this (in pseudocode):
```
NewGraph={}

Original Graph Edges = Edge(Vertex 1,  Vertex 2) and Edge(Vertex 2, Vertex3)

for Edge in Edges:
   NewGraph.Add(Edge.LeftVertex)
   NewGraph.Add(Edge.RightVertex)
```

In the above example, when we examine the second edge, we will encounter Vertex 2 again, because we have already added it when looping through the first edge. This is not an error, because we expect to find certain vertices again while looping if the edges are connected, so I'm not sure if using an exception makes sense here?  

But I definitely agree on constructing an empty graph as the return value!

include/graaflib/graph.tpp
toadkarter: Awesome, love this! Will add.

include/graaflib/graph.tpp
bobluppes: Ah yes, that makes sense then. I was thinking about how the code was also checking whether the passed in graph did not contain the IDs before we do any computation, but this will not be a problem once we construct the graph in the body of the function 👍🏻

include/graaflib/graph.h
```diff
@@ -179,6 +179,16 @@ class graph {
...
 
+  /**
+   * Add a vertex to the graph with a specific ID
+   *
+   * @param  vertex The vertex to be added
+   * @param  id The requested ID for the new vertex
+   * @return vertices_id_t - The ID of the new vertex
+   * @throws id_taken exception - If the relevant ID is already in use
+   */
+  [[nodiscard]] vertex_id_t add_vertex(auto&& vertex, vertex_id_t id);
```
bobluppes, line 190: [very minor]: Since the user already knows the ID of the vertex, it is not necessary to mark the return value as `nodiscard` since the user can safely ignore it.
```suggestion
   * @throws id_taken exception - If the relevant ID is already in use
   */
  vertex_id_t add_vertex(auto&& vertex, vertex_id_t id);
```

include/graaflib/graph.h
```diff
@@ -218,6 +228,18 @@ using directed_graph = graph<VERTEX_T, EDGE_T, graph_type::DIRECTED>;
...
 
+/**
+ * Get transposed version of a given directed graph
+ *
+ * @param  graph The directed graph that is to be transposed
+ * @param  out_transposed_graph An out parameter that will contain the
+ * transposed graph
+ */
+template <typename VERTEX_T, typename EDGE_T>
+void get_transposed_graph(
```
bobluppes, line 239: I think this is left-over code now, as the implementation now lives in `algorithm/utils.h`

include/graaflib/graph.tpp
```diff
@@ -129,12 +129,26 @@ graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::get_neighbors(
 
 template <typename VERTEX_T, typename EDGE_T, graph_type GRAPH_TYPE_V>
 vertex_id_t graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::add_vertex(auto&& vertex) {
-  // TODO: check overflow
-  const auto vertex_id{vertex_id_supplier_++};
+  while (has_vertex(vertex_id_supplier_)) {
+    vertex_id_supplier_++;
+  }
+  const auto vertex_id{vertex_id_supplier_};
```
bobluppes, line 135: [very minor]: Instead of post-increment (`vertex_id_supplier_++`) we could use pre-increment here (`++vertex_id_supplier`). This is slightly faster as post-increment returns the **old** value while pre-increment returns the **new** value. Therefore pre-increment does not need to keep a temporary for the old value.

Since we are not using the return value anyway we can safely do this (arguably very small) optimization :)

include/graaflib/graph.tpp
toadkarter: > Ah yes, that makes sense then. I was thinking about how the code was also checking whether the passed in graph did not contain the IDs before we do any computation, but this will not be a problem once we construct the graph in the body of the function 👍🏻

Great stuff! Those tweaks should now all be pushed. Would you like me to squash all the commits or are you happy for these to stay separate for the time being?

test/graaflib/graph_test.cpp
```diff
@@ -38,6 +38,21 @@ TYPED_TEST(GraphTest, VertexCount) {
...
+  ASSERT_TRUE(graph.has_vertex(specific_id));
+  ASSERT_EQ(graph.get_vertex(specific_id), 30);
+
+  // WHEN - THEN
+  constexpr int colliding_id = 2;
+  const auto vertex_id_4{graph.add_vertex(40)};
+  ASSERT_EQ(graph.vertex_count(), 4);
+  ASSERT_TRUE(graph.has_vertex(colliding_id + 1));
+  ASSERT_EQ(graph.get_vertex(colliding_id), 30);
+  ASSERT_EQ(graph.get_vertex(colliding_id + 1), 40);
```
bobluppes, line 55: In this test section, aren't we also dependent on the implementation specific vertex ID logic since we are checking whether the graph has a vertex with ID `colliding_id + 1`. I remember the original reason for the new `add_vertex` methods was to not be dependent on this implementation detail.

Would it also make sense to get rid of this `WHEN - THEN` section then?

include/graaflib/algorithm/utils.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <graaflib/graph.h>
+
+namespace graaf {
+
+template <typename VERTEX_T, typename EDGE_T>
+directed_graph<VERTEX_T, EDGE_T> get_transposed_graph(
```
bobluppes, line 8: [very minor]: In the codebase we have the habit of splitting definition and implementation. Maybe we can add a `utils.tpp` file for consistency. I would however also be fine to do this in a separate PR as the current one unblocks another PR.

bobluppes (approved):

Just some minor comments, let me know what you think about them and I would be happy to merge :)

As a final thought for future work, I was thinking about what behavior we would expect if I transposed an unconnected graph. Let's say I have graph A with vertices 1, 2, and 3. A only has the directed edge 1 -> 2. Currently if we compute the transpose of A, 3 would not show up in the result set.

What I think we could do in a future implementation is to use the copy constructor for the entire `graph` class. Then we can create a copy, after which we iterate over all edges in the copy. For every edge we can then call `remove_edge` and `add_edge` with the source and target vertices reversed. Wdyt?

toadkarter:
> Thanks for working on this, LGTM!
> 
> Just some minor comments, let me know what you think about them and I would be happy to merge :)
> 
> As a final thought for future work, I was thinking about what behavior we would expect if I transposed an unconnected graph. Let's say I have graph A with vertices 1, 2, and 3. A only has the directed edge 1 -> 2. Currently if we compute the transpose of A, 3 would not show up in the result set.
> 
> What I think we could do in a future implementation is to use the copy constructor for the entire `graph` class. Then we can create a copy, after which we iterate over all edges in the copy. For every edge we can then call `remove_edge` and `add_edge` with the source and target vertices reversed. Wdyt?

Thanks very much for taking the time to review the PR, much appreciated! I will incorporate your commits and do an additional push, squashing the commits at the same time just to make sure that we have one clean commit going into the codebase. I only have one question on these comments which I will put into the relevant thread.

test/graaflib/graph_test.cpp
toadkarter: > In this test section, aren't we also dependent on the implementation specific vertex ID logic since we are checking whether the graph has a vertex with ID `colliding_id + 1`. I remember the original reason for the new `add_vertex` methods was to not be dependent on this implementation detail.
> 
> Would it also make sense to get rid of this `WHEN - THEN` section then?

If I understand correctly, do you mean that this portion of the test might need to be changed down the line because we are relying specifically on how the id is calculated? In that case, I think it's fair to say that this section is too reliant on how `add_vertex` is implemented! I will remove in the next commit

toadkarter:
As for unconnected graphs - that's actually a great point, I haven't thought of that. I would be happy to have a look at that once this and Kosaraju's have gone in!

test/graaflib/graph_test.cpp
bobluppes: Exactly, if we change the underlying ID generation logic this test would break. 

I would say that the newly added method is already sufficiently covered so we can safely remove this part

bobluppes:
> > Thanks for working on this, LGTM!
> > 
> > Just some minor comments, let me know what you think about them and I would be happy to merge :)
> > 
> > As a final thought for future work, I was thinking about what behavior we would expect if I transposed an unconnected graph. Let's say I have graph A with vertices 1, 2, and 3. A only has the directed edge 1 -> 2. Currently if we compute the transpose of A, 3 would not show up in the result set.
> > 
> > What I think we could do in a future implementation is to use the copy constructor for the entire `graph` class. Then we can create a copy, after which we iterate over all edges in the copy. For every edge we can then call `remove_edge` and `add_edge` with the source and target vertices reversed. Wdyt?
> 
> Thanks very much for taking the time to review the PR, much appreciated! I will incorporate your commits and do an additional push, squashing the commits at the same time just to make sure that we have one clean commit going into the codebase. I only have one question on these comments which I will put into the relevant thread. 

Thanks, sounds great! 
As a final remark, we have "squash and merge" enabled, so all main merges are squashed by default. Therefore no need to manually do this on a PR. But also no harm done if you did this already of course :)

bobluppes (approved):

PR #138, opened 2023-10-04
added environment set up link

LinhNguyenLe2109:
@bobluppes  I believe we need to make 2 more things:

- Guide for other OS users like Window and MacOS. If possible you should try and see how you would do it in Window OS. It was really tough for me.
- Easier, straightforward examples in installation guide. I checked the examples, maybe we need something that when we run it prints out something. Or maybe that's unnecessary because people don't use this library to print stuff but to calculate difficult algorithms

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ Graaf is designed with the goal of simplifying graph-related tasks. It offers a 
 
 # Installation
 
+Before you can use the Graaf library, make sure you have your [development environment set up](https://github.com/bobluppes/graaf/wiki/development-setup). Currently, our guide assuming your operating system is Linux. If you're using Windows and IDE such as Visual Studio, Visual Studio Code, feel free to create a guide and submit a pull request showing step-by-step on how you did it on our installation webpage.
+
 The most straightforward way to use the Graaf in your project is to include it as a header-only library. Please take a
 look at the [installation guide](https://bobluppes.github.io/graaf/docs/quickstart/installation) for alternative
 installation methods.
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ analysis, and data mining.
 Graaf: A Lightweight, Header-Only C++20 Graph Library
 
 **Key Features**:
+
 - **Header-only**: No separate compilation or linking required.
 - **Generality**: Supports user-defined vertex and edge classes.
 - **Algorithm Support**: Provides a range of graph algorithms.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,6 +49,8 @@ Graaf is designed with the goal of simplifying graph-related tasks. It offers a
 
 # Installation
 
+Before you can use the Graaf library, make sure you have your [development environment set up](https://github.com/bobluppes/graaf/wiki/development-setup). Currently, our guide assuming your operating system is Linux. If you're using Windows and IDE such as Visual Studio, Visual Studio Code, feel free to create a guide and submit a pull request showing step-by-step on how you did it on our installation webpage.
+
 The most straightforward way to use the Graaf in your project is to include it as a header-only library. Please take a
 look at the [installation guide](https://bobluppes.github.io/graaf/docs/quickstart/installation) for alternative
 installation methods.
```
bobluppes, line 56: Currently we can distinct two types of users:
- Users wanting to integrate `graaf` into their own project
- Developers directly working on the graaf library to improve it (like you 😄 )

This readme and [the quickstart guide](https://bobluppes.github.io/graaf/docs/quickstart/installation) are focused towards the first users, while the wiki is meant for the second.

Therefore I am a little hesitant linking to the wiki here. But maybe we could rephrase it to make this distinction more clear. And we can point to the wiki for troubleshooting.

What do you think?

bobluppes:
> Guide for other OS users like Window and MacOS. If possible you should try and see how you would do it in Window OS. It was really tough for me.

Thanks, this is a good point! I will try it out and create a ticket for it.

> Easier, straightforward examples in installation guide. I checked the examples, maybe we need something that when we run it prints out something. Or maybe that's unnecessary because people don't use this library to print stuff but to calculate difficult algorithms

Good idea, I think printing the dot representation might be helpful to understand how some of the algorithms work. Is this something you would like to work on? I can also create a ticket for it

README.md
LinhNguyenLe2109: Yeah sure, that's a good point, I wasn't sure about this so I'm gonna update this for you

LinhNguyenLe2109:
> > Guide for other OS users like Window and MacOS. If possible you should try and see how you would do it in Window OS. It was really tough for me.
> 
> Thanks, this is a good point! I will try it out and create a ticket for it.
> 
> > Easier, straightforward examples in installation guide. I checked the examples, maybe we need something that when we run it prints out something. Or maybe that's unnecessary because people don't use this library to print stuff but to calculate difficult algorithms
> 
> Good idea, I think printing the dot representation might be helpful to understand how some of the algorithms work. Is this something you would like to work on? I can also create a ticket for it


I would love to take on this issue, but currently I can't get it to work on my computer right now, so I guess I would just deal with the readme first

LinhNguyenLe2109:
@bobluppes I put removed that part and put it in contributing, although we have a link for it already in the wiki page. Let me know if it is necessary, or else you can just close the PR if it's insignificant.

bobluppes:
> > > Guide for other OS users like Window and MacOS. If possible you should try and see how you would do it in Window OS. It was really tough for me.
> > 
> > 
> > Thanks, this is a good point! I will try it out and create a ticket for it.
> > > Easier, straightforward examples in installation guide. I checked the examples, maybe we need something that when we run it prints out something. Or maybe that's unnecessary because people don't use this library to print stuff but to calculate difficult algorithms
> > 
> > 
> > Good idea, I think printing the dot representation might be helpful to understand how some of the algorithms work. Is this something you would like to work on? I can also create a ticket for it
> 
> I would love to take on this issue, but currently I can't get it to work on my computer right now, so I guess I would just deal with the readme first

Sorry to hear that, I will definitely work on improving this

bobluppes (approved):
Nice, I like the addition 👍🏻

PR #139, opened 2023-10-06
[ALGO]Bron-Kerbosch with pivoting

### LINKED ISSUE #95: [ALGO] Bron-Kerbosch Clique Detection
# Bron-Kerbosch
The Bron-Kerbosch algorithm is a backtrack algorithm to find all cliques in an undirected graph. It serves as an essential tool for identifying densely connected subgraphs within a graph and has applications ranging from social network analysis to bioinformatics. 

Implementing this would introduce a new category of algorithms (clique detection) to our library, enhancing its functionality.

See the [wikipedia entry](https://en.wikipedia.org/wiki/Bron%E2%80%93Kerbosch_algorithm) for more details.

## Syntax
The algorithm should have the following syntax:

```c++
/**
 * @brief Finds all cliques in an undirected graph using the Bron-Kerbosch algorithm.
 * 
 * The Bron-Kerbosch algorithm is used for finding all cliques in an undirected graph.
 * A clique is a subset of vertices such that every two distinct vertices are adjacent.
 * This function returns a list of all cliques, each represented as a vector of vertex identifiers.
 * 
 * @tparam V The vertex type of the graph.
 * @tparam E The edge type of the graph.
 * @param graph The graph in which we want to find the cliques.
 * @return A vector of cliques, each represented as a vector of vertex identifiers.
 */
std::vector<std::vector<vertex_id_t>> bron_kerbosch(const graph<V, E, graph_type::UNDIRECTED>& graph);

```

This should live in the `graaf::algorithm` namespace under `include/graaflib/algorithm/clique/bron_kerbosch.h`.

## Definition of Done
This issue is done when:

- [ ] The algorithm is implemented
- [ ] The new function has a javadoc-style comment explaining the interface
- [ ] Appropriate tests are added under `test/graaflib/algorithm/clique/bron_kerbosch_test.cpp`
- [ ] A test coverage of at least 95% is reached
- [ ] A documentation entry is added under `docs/docs/algorithms` under the appropriate category
  - Just adding a short description and the algorithm syntax here is fine
  - See the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation) on how to build the documentation locally
- [ ] The algorithm is added to the list of algorithms in `README.md`

Hromz:
Added algorithm, test and documentation.
In the end I got two implementations of the algorithm, but I decided to stick with the unordered_set implementation. Second one was bitwise approach with ULL implementation, unfortunately it didn't pass the runtime test (I guess I missed something in the implementation). In the future I would like to add bitwise approach and for sparse graph with vertex ordering appraoch.

Looking forward for your comments and suggestions!

closes #95


include/graaflib/algorithm/clique/bron_kerbosch.tpp
```diff
@@ -0,0 +1,97 @@
...
+
+  // Get pivot vertex with most amount of neighbors
+  // to reduce recursion calls
+  auto pivot_vertex = *union_set.begin();
+  for (const auto& vertex : union_set) {
+    if (graph.get_neighbors(pivot_vertex).size() <
+        graph.get_neighbors(vertex).size()) {
+      pivot_vertex = vertex;
+    }
+  }
```
bobluppes, line 49: This works and should be fine to leave as-is. However, if you are up for it this sounds like a great use-case to use C++ 20's [std::ranges::max_element](https://en.cppreference.com/w/cpp/algorithm/ranges/max_element) in combination with a projection.

```c++
const auto pivot_vertex = *std::ranges::max_element(union_set, {}, [&graph](vertex_id_t id) {
  return graph.get_neighbors(id).size();
};
```

include/graaflib/algorithm/clique/bron_kerbosch.tpp
```diff
@@ -0,0 +1,97 @@
...
+
+  // vertices_to_process = candidate_vertices \ N(pivot_vertex)
+  auto vertices_to_process = vertices;
+  for (const auto& vertex : graph.get_neighbors(pivot_vertex)) {
+    vertices_to_process.erase(vertex);
+  }
```
bobluppes, line 55: Same here, might be a nice idea to extract this to a set difference function. i.e.

```c++
std::unordered_set<vertex_id_t> set_difference(const std::unordered_set<vertex_id_t>& lhs, const std::unordered_set<vertex_id_t> rhs) {...}
```

test/graaflib/algorithm/clique/bron_kerbosch_test.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+namespace graaf::algorithm {
+namespace {
+template <typename T>
+struct BronKerbosch : public testing::Test {
+  using graph_t = T;
+};
+
+using graph_types = testing::Types<undirected_graph<int, int>>;
+TYPED_TEST_SUITE(BronKerbosch, graph_types);
```
bobluppes, line 17: [minor]: If we only have one graph type we are testing, there is no real need for a `TYPED_TEST_SUITE`, instead we could just write the type directly in the tests. However, if you do want to test this for different graph types, we now have predefined type collections you can use here in `<utils/fixtures/fixtures.h>`.

test/graaflib/algorithm/clique/bron_kerbosch_test.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+  auto start = std::chrono::high_resolution_clock::now();
+  auto cliques = bron_kerbosch(graph);
+  auto stop = std::chrono::high_resolution_clock::now();
+
+  int expected_time = 900;
+
+  ASSERT_TRUE(expected_time >=
+              duration_cast<std::chrono::milliseconds>(stop - start).count());
+}
```
bobluppes, line 297: Great idea to measure the runtime performance 👍🏻 

Not sure if you saw, but we are also using the google benchmark framework in `/perf`, where we get nice functionalities such as repeated executions of the algorithm after which the average runtime (and memory usage) is reported. There we can also nicely parametrize the benchmark to test for increasingly large graphs.

Such a benchmark might be a better suited place to measure runtime performance of our algorithms than in unit tests. As such I would propose to remove this unit test and to create a follow-up issue in which we can add the benchmark.

bobluppes (commented):
Hi @Hromz, as always great work 🚀 

Hereby some feedback, please let me know what you think.

I really like how you extracted some of the set operations (intersection) into a function. It is super unfortunate that C++ or the stl does not give this for free. In a follow-up, it would be cool if we can extract such set operations to free functions and put them in a central place in the library which we can use as an extension to the stl. Out of scope for this PR but just some food for thought :)

Hromz:
Hey @bobluppes! Thanks for your comments, with each PR I'm discovering something new. :)  I corrected the code according to your comments. Renamed folder from "Clique" to "Clique-detection". Deleted the runtime test and added two more tests.

README.md
```diff
@@ -92,24 +92,27 @@ Algorithms implemented in the Graaf library include the following. For more info
...
-   - Dijkstra
-   - Bellman-Ford
-5. [**Strongly Connected Components Algorithms**](https://bobluppes.github.io/graaf/docs/category/strongly-connected-components):
-   - Tarjan's Strongly Connected Components
+    - A\* search
+    - BFS-Based Shortest Path
+    - Dijkstra
+    - Bellman-Ford
+5. [**Strongly Connected Components Algorithms
+   **](https://bobluppes.github.io/graaf/docs/category/strongly-connected-components):
```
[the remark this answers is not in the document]
Hromz: Hmm, think i pressed auto alignment by mistake. Corrected it.

bobluppes (approved):
Thanks for addressing, looks excellent 🎉

Issue #115, opened 2023-09-30
[ALGO] Kosaraju's Algorithm


bobluppes:
# Kosaraju's Algorithm
Kosaraju's algorithm identifies the strongly connected components (SCCs) in a directed graph. A strongly connected component is a subgraph in which there is a path from every vertex to every other vertex. The algorithm uses two depth-first searches to identify these components.

When possible, it would be nice to see if we can reuse the existing DFS functionality in `algorithm/graph_traversal/depth_first_search.h`.

For more details, see the [wikipedia entry](https://en.wikipedia.org/wiki/Kosaraju%27s_algorithm).

## Syntax
The algorithm should have the following syntax:

```c++
/**
 * @brief Finds strongly connected components in a directed graph using Kosaraju's algorithm.
 *
 * Kosaraju's algorithm identifies strongly connected components (SCCs) in a directed graph.
 * The algorithm uses two depth-first searches to discover these components. 
 *
 * @tparam V The vertex type of the graph.
 * @tparam E The edge type of the graph.
 * @param graph The input directed graph.
 * 
 * @return A std::vector of vectors, each of which contains the vertex IDs forming
 *         a strongly connected component.
 */
template <typename V, typename E>
std::vector<std::vector<vertex_id_t>> kosarajus_strongly_connected_components(
    const graph<V, E, graph_type::DIRECTED>& graph
);
```

This should live in the `graaf::algorithm` namespace under `include/graaflib/algorithm/strongly_connected_components/kosarajus.h`.

## Definition of Done
This issue is done when:

- [ ] The algorithm is implemented
- [ ] The new function has a javadoc-style comment explaining the interface
- [ ] Appropriate tests are added under `test/graaflib/algorithm/strongly_connected_components/kosarajus_test.cpp`
- [ ] A test coverage of at least 95% is reached
- [ ] A documentation entry is added under `docs/docs/algorithms` under the appropriate category
  - Just adding a short description and the algorithm syntax here is fine
  - See the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation) on how to build the documentation locally
- [ ] The algorithm is added to the list of algorithms in `README.md`

toadkarter:
Hi there - happy to take a look at this one if it's still available.

bobluppes:
> Hi there - happy to take a look at this one if it's still available. If this issue could be assigned to me that would be great.

Hi, of course! Looking forward to your contribution!

toadkarter:
[...]

PR #141, opened 2023-10-08
[ALGO] Kosaraju's Algorithm

toadkarter:
Implements #115

&#x2611;  The algorithm is implemented
&#x2611;  The new function has a javadoc-style comment explaining the interface
&#x2611; Appropriate tests are added under `test/graaflib/algorithm/strongly_connected_components/kosarajus_test.cpp` 
&#x2611; A test coverage of at least 95% is reached
&#x2611; A documentation entry is added under `docs/docs/algorithms` under the appropriate category. Just adding a short description and the algorithm syntax here is fine. See the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation) on how to build the documentation locally
&#x2611; The algorithm is added to the list of algorithms in `README.md`

I should flag here that I decided to use a simple recursive DFS - the one that was already in the library seemed (unless I am mistaken) more geared towards searching through edges as opposed to vertices. However, if the library implementation of DFS is required please do let me know and I can work it into the existing code.

### DIFF
--- a/include/graaflib/algorithm/strongly_connected_components/common.h
+++ b/include/graaflib/algorithm/strongly_connected_components/common.h
@@ -1,0 +1,11 @@
+#pragma
+
+#include <graaflib/types.h>
+
+#include <vector>
+
+namespace graaf::algorithm {
+
+using sccs_t = std::vector<std::vector<vertex_id_t>>;
+
+}  // namespace graaf::algorithm
--- a/include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
+++ b/include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
@@ -17,12 +17,11 @@ void do_visit_vertex(const vertex_id_t vertex_id,
                      std::stack<vertex_id_t>& stack,
                      std::unordered_set<vertex_id_t>& seen_vertices) {
   if (!seen_vertices.contains(vertex_id)) {
-    stack.push(vertex_id);
     seen_vertices.insert(vertex_id);
-
     for (const auto neighbour : graph.get_neighbors(vertex_id)) {
       do_visit_vertex(neighbour, graph, stack, seen_vertices);
     }
+    stack.push(vertex_id);
   }
 }
 
--- a/include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
+++ b/include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
@@ -9,6 +9,41 @@
 
 namespace graaf::algorithm {
 
+namespace {
+
+template <typename V, typename E>
+void do_visit_vertex(const vertex_id_t vertex_id,
+                     const directed_graph<V, E>& graph,
+                     std::stack<vertex_id_t>& stack,
+                     std::unordered_set<vertex_id_t>& seen_vertices) {
+  if (!seen_vertices.contains(vertex_id)) {
+    stack.push(vertex_id);
+    seen_vertices.insert(vertex_id);
+
+    for (const auto neighbour : graph.get_neighbors(vertex_id)) {
+      do_visit_vertex(neighbour, graph, stack, seen_vertices);
+    }
+  }
+}
+
+template <typename V, typename E>
+void make_strongly_connected_component_from_vertex(
+    const vertex_id_t vertex, const directed_graph<V, E>& transposed_graph,
+    std::vector<vertex_id_t>& scc,
+    std::unordered_set<vertex_id_t>& seen_vertices) {
+  if (!seen_vertices.contains(vertex)) {
+    seen_vertices.insert(vertex);
+    scc.push_back(vertex);
+
+    for (const auto neighbour : transposed_graph.get_neighbors(vertex)) {
+      make_strongly_connected_component_from_vertex(neighbour, transposed_graph,
+                                                    scc, seen_vertices);
+    }
+  }
+}
+
+}  // namespace
+
 template <typename V, typename E>
 std::vector<std::vector<vertex_id_t>> kosarajus_strongly_connected_components(
     const directed_graph<V, E>& graph) {
@@ -21,21 +56,9 @@ std::vector<std::vector<vertex_id_t>> kosarajus_strongly_connected_components(
   std::stack<vertex_id_t> stack{};
   std::unordered_set<vertex_id_t> seen_vertices{};
 
-  std::function<void(vertex_id_t)> visit_vertex;
-  visit_vertex = [&](vertex_id_t vertex) {
-    if (!seen_vertices.contains(vertex)) {
-      stack.push(vertex);
-      seen_vertices.insert(vertex);
-
-      for (const auto neighbour : graph.get_neighbors(vertex)) {
-        visit_vertex(neighbour);
-      }
-    }
-  };
-
   for (const auto& [vertex_id, vertex] : graph.get_vertices()) {
     if (!seen_vertices.contains(vertex_id)) {
-      visit_vertex(vertex_id);
+      do_visit_vertex(vertex_id, graph, stack, seen_vertices);
     }
   }
 
@@ -44,22 +67,11 @@ std::vector<std::vector<vertex_id_t>> kosarajus_strongly_connected_components(
   directed_graph<V, E> transposed_graph = get_transposed_graph(graph);
   std::vector<vertex_id_t> scc{};
 
-  std::function<void(vertex_id_t)> strong_connect;
-  strong_connect = [&](vertex_id_t vertex) {
-    if (!seen_vertices.contains(vertex)) {
-      seen_vertices.insert(vertex);
-      scc.push_back(vertex);
-
-      for (const auto neighbour : transposed_graph.get_neighbors(vertex)) {
-        strong_connect(neighbour);
-      }
-    }
-  };
-
   while (!stack.empty()) {
     if (!seen_vertices.contains(stack.top())) {
       scc.clear();
-      strong_connect(stack.top());
+      make_strongly_connected_component_from_vertex(
+          stack.top(), transposed_graph, scc, seen_vertices);
       sccs.push_back(scc);
     }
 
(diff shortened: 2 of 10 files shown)


include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
```diff
@@ -0,0 +1,72 @@
...
+  visit_vertex = [&](vertex_id_t vertex) {
+    if (!seen_vertices.contains(vertex)) {
+      stack.push(vertex);
+      seen_vertices.insert(vertex);
+
+      for (const auto neighbour : graph.get_neighbors(vertex)) {
+        visit_vertex(neighbour);
+      }
+    }
+  };
```
bobluppes, line 34: In [Kosaraju](https://en.wikipedia.org/wiki/Kosaraju%27s_algorithm), should we not add (prepend) the current vertex to the stack (L in the wikipedia article) **after** visiting its neighbors?

include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
```diff
@@ -0,0 +1,72 @@
...
+  strong_connect = [&](vertex_id_t vertex) {
+    if (!seen_vertices.contains(vertex)) {
+      seen_vertices.insert(vertex);
+      scc.push_back(vertex);
+
+      for (const auto neighbour : transposed_graph.get_neighbors(vertex)) {
+        strong_connect(neighbour);
+      }
+    }
+  };
```
bobluppes, line 57: Same here, what do you think about extracting this to a separate function?

include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
```diff
@@ -0,0 +1,72 @@
...
+
+  for (const auto& [vertex_id, vertex] : graph.get_vertices()) {
+    if (!seen_vertices.contains(vertex_id)) {
+      visit_vertex(vertex_id);
+    }
+  }
+
+  seen_vertices.clear();
+
+  directed_graph<V, E> transposed_graph = get_transposed_graph(graph);
```
bobluppes, line 44: [very minor]: To save some typing (and to be `const`-correct) you could do
```suggestion

  const auto transposed_graph = get_transposed_graph(graph);
```

test/graaflib/algorithm/strongly_connected_components/kosaraju_test.cpp
```diff
@@ -0,0 +1,53 @@
+#include <graaflib/algorithm/strongly_connected_components/kosaraju.h>
+#include <gtest/gtest.h>
+
+namespace graaf::algorithm {
+
+TEST(KosarajuTest, EqualOutputSCCS) {
+  using sccs_t = std::vector<std::vector<vertex_id_t>>;
```
bobluppes, line 7: It would be nice to have some more tests here, but I would be fine with doing this in a follow-up. Maybe you could reuse some of the predefined graph scenarios in `test/utils/scenarios/scenarios.h`.

include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
bobluppes: I suspect the algorithm might fail for the following graph:
```c++
  directed_graph<int, int> graph{};
  const auto d = graph.add_vertex(4);
  const auto a = graph.add_vertex(1);
  const auto b = graph.add_vertex(2);
  const auto c = graph.add_vertex(3);


  graph.add_edge(a, b, 1);
  graph.add_edge(b, c, 1);
  graph.add_edge(c, a, 1);
  graph.add_edge(b, d, 1);
```

Where we would expect SCCs {A, B, C} and {D}. Might be a nice test to add.

test/graaflib/algorithm/strongly_connected_components/kosaraju_test.cpp
bobluppes, line 7: Nice using-declaration 👍🏻 Might also be cool to move this to the source file!

bobluppes (changes requested):
Thanks for working on this!
Looks quite good already, was just thinking about a possible issue regarding the order in which we add vertices to the set in the first DFS.

include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
```diff
@@ -0,0 +1,72 @@
...
+#include <iostream>
+#include <stack>
+
+namespace graaf::algorithm {
+
+template <typename V, typename E>
+std::vector<std::vector<vertex_id_t>> kosarajus_strongly_connected_components(
+    const directed_graph<V, E>& graph) {
+  std::vector<std::vector<vertex_id_t>> sccs;
+
```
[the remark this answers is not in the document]
toadkarter: Thanks, I wasn't aware of that! Will make that tweak.

include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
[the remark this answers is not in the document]
toadkarter: Interesting, that sounds good! Could I ask why we would want to wrap it into another namespace here as opposed to just having it as a standalone function in the `.tpp` file - does this just signify that only functions in the `.tpp` file are able to access it and that users of the header won't be able to call the helper? 

Also, with situations like this would we still do a definition of the helper function in the header and have the definition in the `.tpp`?

toadkarter: I think that's right, that's my bad - will make this tweak!

include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
toadkarter: Sounds good

include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
toadkarter: Perfect, will add

test/graaflib/algorithm/strongly_connected_components/kosaraju_test.cpp
toadkarter: Awesome - does it make sense to pop it into `types`? Or to make a `types` header for strongly connected components, as I know Tarjan would also benefit from this?

toadkarter: That makes sense - I think based on the code coverage I also could add a test on what happens if a blank graph is passed in. 

I will do up the tests for the ones in `scenarios.h`, are there any other ones that you possibly had in mind?

toadkarter:
> Thanks for working on this! Looks quite good already, was just thinking about a possible issue regarding the order in which we add vertices to the set in the first DFS.

Thanks very much for the review! All of these suggestions look great and I will try to implement them when back from work today.

test/graaflib/algorithm/strongly_connected_components/kosaraju_test.cpp
bobluppes: The `types.h` file is meamt for project wide types such as `vertex_id_t`. But I like the idea of sharing the typedef with Tarjan'!

What do you think about adding a `strongly_connected_components/common.h`?

bobluppes: Sounds good to me! I would also add the graph setup for which the algorithm currently fails. Any follow up scenarios we could also do in a follow-up!

include/graaflib/algorithm/strongly_connected_components/kosaraju.tpp
bobluppes: No worries, thanks a lot!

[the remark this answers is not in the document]
bobluppes: You are correct that adding it to the anonymous namespace will make it unavailable to any other compilation units. So we are not polluting the namespace of whoever includes this file.

In general I would restrict the scope of any helper functions as much as possible and only make it available on the interface if absolutely needed

test/graaflib/algorithm/strongly_connected_components/kosaraju_test.cpp
toadkarter: > The `types.h` file is meamt for project wide types such as `vertex_id_t`. But I like the idea of sharing the typedef with Tarjan'!
> 
> What do you think about adding a `strongly_connected_components/common.h`?

I think that makes a lot of sense, and any file referencing Tarjan / Kosaraju / etc. will have access to `common` because it will be `include`d in those headers. Will give it a go!

include/graaflib/algorithm/strongly_connected_components/common.h
```diff
@@ -0,0 +1,11 @@
+#pragma
+
+#include <graaflib/types.h>
+
+#include <vector>
+
+namespace graaf::algorithm {
+
+using sccs_t = std::vector<std::vector<vertex_id_t>>;
```
bobluppes, line 9: 👍🏻 very nice that we now have this. Also thanks for updating Tarjan!

test/graaflib/algorithm/strongly_connected_components/kosaraju_test.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  // THEN
+  ASSERT_TRUE(are_set_vectors_equal(sccs, expected_sccs));
+}
+
+TEST(KosarajuTest, DisconnectedGraphScenario) {
+  // GIVEN
+  const auto [graph, vertex_ids] =
+      utils::scenarios::create_disconnected_graph_scenario<
+          directed_graph<int, int>>();
```
bobluppes, line 141: Also nice that we are now using these scenario's in the tests! I would like to start using them more throughout the codebase

bobluppes (approved):
Looks excellent 🎉

PR #152, opened 2023-10-11
Feat : Added Documentation for Welsh Powell Algorithm

### LINKED ISSUE #146: [DOCS] Document Welsh-Powell Algorithm
# Documentation Welsh-Powell
The goal of this issue if the improve the documentation of an existing algorithm. This documentation should go into the algorithms section of our [docusaurus docs](https://bobluppes.github.io/graaf/docs/algorithms/intro).

More detail on how to build the documentation locally can be found on the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation).

A new documentation page should be created under `docs/docs/algorithms/coloring`.

The corresponding algorithm is called `Welsh-Powell` and can be found under `include/graaflib/algorithm/coloring/welsh_powell.h`.

## Documentation Contents
The documentation entry should adhere to the following template:

```markdown
# [ALGORITHM_NAME]
- A short description of the algorithm, what does it do and in which use cases is it used.
- What are the limitations of the algorithm, does it work on both directed and undirected graphs? Does it consider edge weights, and if yes, does it support negative weights?
- Link to the wikipedia entry on the algorithm.

## Syntax
- A code block with the syntax of the algorithm. This should now include any javadoc comments.
- An explanation of the parameters (including template parameters) and the return type.

## (Optional) See also
- If there are similar algorithms, or we have a slightly different version of the same algorithm, you can link it here.
- If we have an example (found on the example page of the docs) which uses this algorithm, you can link it here.
```

GouraV-Chatterjee:
Fixes #146

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,9 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+
+  - package-ecosystem: "npm"
+    # Look for `package.json` and `lock` files in the `root` directory
+    directory: "/docs"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,28 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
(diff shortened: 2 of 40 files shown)


bobluppes (approved):
Thanks, LGTM 🚀

PR #153, opened 2023-10-11
Create Welsh-Powell .md

spideyWebb:
Welsh-Powell Graph Coloring Algorithm

### DIFF
--- a/docs/docs/algorithms/coloring/Welsh-Powell .md
+++ b/docs/docs/algorithms/coloring/Welsh-Powell .md
@@ -1,0 +1,40 @@
+# Welsh-Powell Graph Coloring Algorithm
+
+The Welsh-Powell algorithm is a graph coloring algorithm used to assign colors to the vertices of an undirected graph such that no two adjacent vertices have the same color. This algorithm was introduced by Welsh and Powell in 1967 and is a simple and efficient approach for solving graph coloring problems.
+
+## Algorithm
+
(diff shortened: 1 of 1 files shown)


docs/docs/algorithms/coloring/Welsh-Powell .md
bobluppes, line 1: It seems like this filename is causing the CI to fail. Removing the whitespace should work. In general we also have the convention to use lower case filenames, so maybe you could change it to `welsh-powell.md`.

You could check if it works locally by:
```bash
cd docs
yarn install && yarn build
```

See [this wiki guide](https://github.com/bobluppes/graaf/wiki/development-setup#documentation) for more details.

docs/docs/algorithms/coloring/Welsh-Powell .md
```diff
@@ -0,0 +1,40 @@
...
+            if neighbor in colors:
+                usedColors.add(colors[neighbor])
+        
+        # Find the lowest available color
+        for color in range(len(usedColors) + 1):
+            if color not in usedColors:
+                colors[vertex] = color
+                break
+    
+    return colors
```
bobluppes, line 40: I think your explanation above already works quite well, so I would also be fine with leaving out the pseudocode. However, if you decide to keep it then could you please add the closing tag (```) of the code block as well?

bobluppes (changes requested):
Thanks for working on this, looks almost ready to merge

PR #157, opened 2023-10-12
[BENCH] Bron-Kerbosch clique detection

Hromz:
Added a benchmark for the Bron-Kerbosch algorithm. This implementation consists of sparse, dense, custom, and random graphs. Additionally, I added ms. instead of nn and BigO notation. If you think that is inappropriate here, let me know and I will remove it.


perf/graaflib/bron_kerbosch_benchmark.cpp
```diff
@@ -0,0 +1,245 @@
...
+static void bron_kerbosh_graph_triangle_cliques(benchmark::State& state) {
+  const auto number_of_edges{static_cast<size_t>(state.range(0))};
+
+  graaf::undirected_graph<int, int> graph{};
+
+  // We create enough vertices to construct the requested number of edges
+  const auto number_of_vertices{number_of_edges + 1};
+  const auto vertices{create_vertices(graph, number_of_vertices)};
+
+  size_t clique_size = 4;
```
bobluppes, line 101: Should the `clique_size` not be equal to 3 if we want to make triangle cliques?

perf/graaflib/bron_kerbosch_benchmark.cpp
```diff
@@ -0,0 +1,245 @@
...
+static void bron_kerbosh_graph_large_cliques(benchmark::State& state) {
+  const auto number_of_edges{static_cast<size_t>(state.range(0))};
+
+  graaf::undirected_graph<int, int> graph{};
+
+  // We create enough vertices to construct the requested number of edges
+  const auto number_of_vertices{number_of_edges + 1};
+  const auto vertices{create_vertices(graph, number_of_vertices)};
+
+  size_t clique_size = 25;
```
bobluppes, line 151: We could also do this in a follow-up, but looking at these parameters it looks like we can pull out the `clique_size` as a second parameter. Then we might be able to consolidate a few of these benchmarks. I.e.

```c++
BENCHMARK(bron_kerbosh_connected_cliques)
    ->Ranges({{100, 10000}, {3, 100}})  // total number of edges, clique size
    ->Unit(benchmark::kMillisecond)
    ->Complexity();
```

perf/graaflib/bron_kerbosch_benchmark.cpp
```diff
@@ -0,0 +1,245 @@
...
+#include <random>
+#include <vector>
+
+namespace {
+
+// Generating random number of vertices (1- 30) for a clique
+std::random_device dev;
+std::mt19937 rng(dev());
+std::uniform_int_distribution<std::mt19937::result_type> random_clique_size(1,
+                                                                            30);
```
bobluppes, line 14: I am always a little hesitant adding such non deterministic benchmarks, as it makes comparing results difficult. It might still be nice to have since we use it to determine the runtime complexity as well, just something to be aware of.

bobluppes (changes requested):
Looks quite good already, a very extensive set of benchmarks 👍🏻

perf/graaflib/bron_kerbosch_benchmark.cpp
```diff
@@ -0,0 +1,245 @@
...
+  // Constructing a clique
+  for (size_t i{start_vertex}; i < end_vertex; ++i) {
+    for (size_t j{i + 1}; j < end_vertex; ++j) {
+      clique.emplace_back(vertices[i], vertices[j]);
+    }
+  }
+
+  for (auto& edge_t : clique) {
+    graph.add_edge(edge_t.first, edge_t.second, 1);
+  }
```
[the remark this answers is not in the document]
Hromz: Oh my bad, I overconfused myself.

perf/graaflib/bron_kerbosch_benchmark.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+}  // namespace
+
+static void bron_kerbosh_two_vertices_cliques(benchmark::State& state) {
```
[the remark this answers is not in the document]
Hromz: Agreed, that definitely will be a better explanation of what code does.

perf/graaflib/bron_kerbosch_benchmark.cpp
```diff
@@ -0,0 +1,245 @@
...
+  const auto vertices{create_vertices(graph, number_of_vertices)};
+
+  for (auto _ : state) {
+    for (size_t i{0}; i < number_of_edges; ++i) {
+      graph.add_edge(vertices[i], vertices[i + 1], 1);
+    }
+  }
+
+  graaf::algorithm::bron_kerbosch(graph);
+  state.SetComplexityN(state.range(0));
```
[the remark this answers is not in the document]
Hromz: Alright!

perf/graaflib/bron_kerbosch_benchmark.cpp
Hromz: My bad with naming, basically,  it's a triangle with a dot inside (4 vertex clique).

perf/graaflib/bron_kerbosch_benchmark.cpp
Hromz: If you want I would like to add this in currecnt PR!

bobluppes (approved):
Thanks for working on this, LGTM

PR #164, opened 2023-10-20
[BENCH] DFS traversal

Hromz:
Added benchmark for DFS traversal. The benchmark contains several graph paths, such as binary, matrix, and cone-like graphs. I didn't add tests for undirected graphs because I assumed that would be excessive, but if it's not, I would like to add! This is not 100% ready for PR; I just want to hear your opinion about it.

Thanks in advance. :)


Hromz:
Added benchmark for undirected graph

Issue #10, opened 2023-05-20
[DOCS] Create Example Sections


bobluppes:
# Create Example Sections
This is an open-ended issue and we welcome multiple PRs.

I would like to add a few example sections to the documentation. Each example should be self-contained and should demonstrate a potential usage of the Graaf library.

If you are interested, you can pick an example from the proposals below, or (even better) suggest your own example. The scope of a PR is restricted to a single example.

## Definition of Done
Each example should contain the following:

- [ ] A complete and self-contained example under `examples/EXAMPLE_NAME/main.cpp`.
  - To verify that the example compiles and runs.
  - This can be used to generate any graphics (via dot files) for the public documentation.
  - See the [dot serialization example](https://github.com/bobluppes/graaf/blob/main/examples/dot_serialization/main.cpp) for inspiration.
- [ ] An entry on our [public documentation](https://bobluppes.github.io/graaf/docs/examples/Intro).
  - A short description of the example, what is the use case and why is it typical.
  - A list of the Graaf algorithms and data structures which are demonstrated in the example. Here you can link to the respective [algorithms page](https://bobluppes.github.io/graaf/docs/algorithms/intro) in the documentation.
  - Code pointers (in the form of code blocks) to the important sections of the example. It is not necessary to cover the entirety of the added `main.cpp`, but the most important concepts should be captured.
  - Non-trivial parts in the code blocks should be explained in-text.
  - Graphics (i.e. in the form of visualized dot files) can be used. Any graphics should go under `docs/static/img/examples/.
  - Instructions on how to build the documentation locally can be found on the [wiki](https://github.com/bobluppes/graaf/wiki/development-setup#documentation).
  - An example entry can be found [here](https://bobluppes.github.io/graaf/docs/examples/example-basics/dot-serialization).

## Proposed Examples
- [x] Dot serialization
- [x] Shortest path (directed graph)
- [x] Shortest path (undirected graph) [@Hromz]
- [ ] Cycle detection [@Hromz] 
- [ ] Social network use case
- [ ] Minimal CNN implementation
- [ ] ...

Hromz:
Hey @bobluppes, i would like to work on examples for documentation. Can i try?

bobluppes:
Hi @Hromz, of course, looking forward to your contribution!

[...]

PR #167, opened 2023-10-23
added social_network example

Poornima2003:
Issue #10 
I have added the example for social_network section

### DIFF
--- a/examples/social_network/main.cpp
+++ b/examples/social_network/main.cpp
@@ -1,0 +1,35 @@
+
+#include <iostream>
+#include <graaflib/algorithm/social_network.h>
+
+
+int main() {
+    // Create a graph to represent the social network
+    Graaf::Graph socialNetwork;
+
+    // Add nodes for users in the social network
+    Graaf::Node userAlice = socialNetwork.addNode("Alice");
+    Graaf::Node userBob = socialNetwork.addNode("Bob");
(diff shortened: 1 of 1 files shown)


examples/social_network/main.cpp
```diff
@@ -0,0 +1,35 @@
+
+#include <iostream>
+#include <graaflib/algorithm/social_network.h>
+
+
+int main() {
+    // Create a graph to represent the social network
+    Graaf::Graph socialNetwork;
```
joweich, line 8:
```suggestion
    graaf::undirected_graph<V, E> graph{};
```
You need to specify the graph type here. I think `undirected` is reasonable for social networks. The `V` and `E` should be replaced with more reasonable types according to the specific use case

joweich (changes requested):
Hey @Poornima2003, great to see your contribution!
However, the references/functions/types you're using aren't part of our codebase under these names. You can check the other examples in this repo. 
I assume you used an LLM for inspiration (which is totally fine). I'd suggest to feed it with parts of our codebase to increase the models fit.
In any case, the code needs to compile and run successfully on your local machine and in our CI. Let me know if you need help! 🙂

bobluppes:
Hi @Poornima2003, closing this for now. Feel free to re-open if you want to continue working on it

Issue #87, opened 2023-09-04
[ALGO] Graph Isomorphism


bobluppes:
# Graph Isomorphism
In graph theory, an isomorphism between graphs G and H is a structure-preserving bijection between their vertex sets that also preserves adjacency. If such an isomorphism exists, the graphs are considered isomorphic, denoted as G≃HG≃H. Determining graph isomorphism efficiently is a major open question in computer science, known as the Graph Isomorphism problem.

The proposal is to implement the VF2 algorithm. The VF2 algorithm is a backtracking algorithm that provides a generic way to perform both subgraph and graph isomorphism tests. It is designed to be efficient and can handle both directed and undirected graphs. The algorithm is implemented in a way that allows for attributes to be associated with vertices and edges, and for those attributes to be considered as part of the isomorphism test if needed.

See the [wikipedia entry](https://en.wikipedia.org/wiki/Subgraph_isomorphism_problem) for more details.

## Syntax
The algorithm should have the following syntax:

```c++
/**
 * @brief Checks if two graphs are isomorphic.
 *
 * This function checks whether two given graphs are isomorphic, meaning
 * they are structurally identical up to vertex and edge relabeling.
 * If the graphs are isomorphic, the function returns an optional containing
 * a mapping from the vertices of the first graph to the vertices of the second graph.
 * If the graphs are not isomorphic, the function returns std::nullopt.
 *
 * @tparam V The vertex type of the graphs.
 * @tparam E The edge type of the graphs.
 * @tparam T The graph type of the graphs (directed or undirected).
 * @param lhs The first graph to compare.
 * @param rhs The second graph to compare.
 * @return An optional containing a mapping from the vertices of lhs to those of rhs if isomorphic; otherwise std::nullopt.
 */
template <
    typename V, typename E, graph_type T
>
std::optional<std::unordered_map<vertex_id_t, vertex_id_t>> check_isomorphism(
    const graph<V, E, T>& lhs,
    const graph<V, E, T>& rhs
);
```

This should live in the `graaf::algorithm` namespace under `include/graaflib/algorithm/graph_isomorphism.h`.

## Definition of Done
This issue is done when:

- [ ] The algorithm is implemented
- [ ] The new function has a javadoc-style comment explaining the interface
- [ ] Appropriate tests are added under `test/graaflib/graph_isomorphism_test.cpp`
- [ ] A test coverage of at least 95% is reached
- [ ] A documentation entry is added under `docs/docs/algorithms` under the appropriate category
[...]

PR #172, opened 2023-11-04
VF2 graph isomorphism revised PR for issue #87, requesting review 

sracha4355:
Changes I've made:
- removed unnecessary functions from .h, .h contains only the function interface that will be used by the end-user
- added comments to my code and tests
- added some new tests to test isomorphism
- formatted the code with clang-format using style Google
I appreciate any constructive criticism you guys can offer. As for the comments, it is hard to tell the full story of the algorithm given it's complex nature, so in addition to my comments I linked the research paper if any devs are still confused. 
Thank you guys!


sracha4355:
Most of the added lines in PR are from cmake files i forgot to push because I rebuilt the library.

sracha4355:
@bobluppes @joweich could I get a workflow approval?

sracha4355:
@joweich can I get a review

include/graaflib/algorithm/graph_isomorphism.tpp
```diff
@@ -0,0 +1,740 @@
...
+  }
+  for (const auto& vertex : rhs.get_vertices()) {
+    std::size_t number_of_conn = rhs.get_neighbors(vertex.first).size();
+    rhs_node_degrees.push_back(number_of_conn);
+  }
+  std::sort(lhs_node_degrees.begin(), lhs_node_degrees.end());
+  std::sort(rhs_node_degrees.begin(), rhs_node_degrees.end());
+  for (int i = 0; i < lhs_node_degrees.size(); i++)
+    if (lhs_node_degrees[i] != rhs_node_degrees[i]) return false;
+  return true;
```
joweich, line 737:
```suggestion
  auto compute_and_sort_degrees = [](const graph<V, E, T>& graph) {
      std::vector<std::size_t> node_degrees;
      for (const auto& vertex : graph.get_vertices()) {
          std::size_t number_of_conn = graph.get_neighbors(vertex.first).size();
          node_degrees.push_back(number_of_conn);
      }
      std::sort(node_degrees.begin(), node_degrees.end());
      return node_degrees;
  };

  const auto lhs_node_degrees = compute_and_sort_degrees(lhs);
  const auto rhs_node_degrees = compute_and_sort_degrees(rhs);

  return lhs_node_degrees == rhs_node_degrees;
```
I think using a lambda would be more concise here

include/graaflib/algorithm/graph_isomorphism.tpp
```diff
@@ -0,0 +1,740 @@
...
+
+  if (valid_mapping_found) {
+    vertex_mapping mapping;
+    // map vertex ids from lhs to rhs
+    for (int i = 0; i < state->lhs_core.size(); i++)
+      mapping[state->lhs_reverse_node_order.at(i)] =
+          state->rhs_reverse_node_order.at(state->lhs_core[i]);
+    delete state;
+    std::optional<vertex_mapping> _optional = mapping;
+    return _optional;
```
joweich, line 187:
```suggestion
    return mapping;
```
This should be possible with implicit template deduction.

include/graaflib/algorithm/graph_isomorphism.tpp
```diff
@@ -0,0 +1,740 @@
...
+  for (const auto& element : set_A) {
+    if (set_B.count(element) > 0) intersection.insert(element);
+  }
+  return std::move(intersection);
+}
+
+// a lookahead rule to see if the current partial vertex_mapping can lead to a
+// complete isomorphic vertex_mapping
+template <typename V, typename E, graph_type T>
+bool r_new(const graph<V, E, T>& lhs, const graph<V, E, T>& rhs,
```
joweich, line 670: this function is very repetitive and its name should be more expressive

include/graaflib/algorithm/graph_isomorphism.tpp
```diff
@@ -0,0 +1,740 @@
...
+    }
+  }
+  return false;
+}
+
+void update_terminal_set(const node_set& vertices, terminal_set& _terminal_set,
+                         size_t depth, size_t& _terminal_set_length) {
+  for (const auto& node : vertices) {
+    if (_terminal_set[node] == -1) {
+      _terminal_set[node] = int(depth);
```
joweich, line 226: why is depth defined as size_t when casted to int here?

joweich (commented):
Hey @sracha4355, great to see you incorporated the feedback from last time!
I'll be a bit more specific with my feedback this time compared to the guardrails last time :slightly_smiling_face: 

- The logic itself is quite hard to follow because the level of abstraction changes often and some sections are quite repetitive. I think we should put some thought into introducing some object orientation to be more concise about the dependencies. `terminal_set` or `vf2_vertex` would be candidates to consider introducing as structs/classes
- There are several return statements that lack test coverage. There should be tests if the returns are necessary, otherwise we should remove them
- You sometimes allocate memory with "naked" new statements. Most often,  it's better (i.e. more robust against memory leaks) to use smart pointers. Won't go into detail here, but check e.g. [this](https://en.cppreference.com/book/intro/smart_pointers)

I did not review every function, but I hope the bullets above and the inline-comments provide some useful advice already! :slightly_smiling_face:

sracha4355:
@joweich thank you for your comments, I'll take a look at each one and refactor the code!

sracha4355:
Hey guys, I got really busy with school, but now that I am on break I would like to refactor the code and get my code merged. Can we reopen the PR?

PR #195, opened 2024-02-08
[TOOL] Clean Python dependencies

joweich:
Removes unnecessary Python dependencies. Will prevent dependabot from opening PRs like https://github.com/bobluppes/graaf/pull/188 and https://github.com/bobluppes/graaf/pull/191.


bobluppes (approved):
Thanks for opening, LGTM!

PR #217, opened 2024-06-11
CMake Targets for FetchContent

FlorianXXIV:
I tried to use FetchContent for CMake to install graaflib but it didn't work, after some testing I decided to create a fork.

I made a new interface with add_library and added the include directory over target_include_directories, when using FetchContent with these modifications, including graaflib worked.

I also made sure I could build and test the project so that this didn't break stuff, and I got a build error telling me that the tarjan test used std::sort as an unknown target, I added an include for <algorithm> in tarjan.h to make the project build and the tests ran at 100%

I'm not really acquainted with how you properly write CMakeList files so if there are any issues feel free to tell me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 build/
 cmake-modules/
 venv
+cmake-build-debug
 
 # IDE
 .vscode/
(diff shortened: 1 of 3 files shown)


FlorianXXIV:
I forgot to add a lable when creating the request and now It seems I can't do that afterwards. Sorry about that

bobluppes (approved):
Hi @FlorianXXIV, thanks for adding this, LGTM!

In the future I think we should add a test in the release CI to ensure that the project can be included using FetchContent. In general I would see this as a larger topic as the installation workflow for users could be streamlined 😅 

This is a really nice step in that direction though! Please let me know if you continue to encounter issues.

PR #221, opened 2024-06-19
[CI] Re-add clang to matrix build

joweich:
Reverts #169. The root cause in the respective runner has been fixed recently (https://github.com/actions/runner-images/issues/9679).


bobluppes (approved):
Nice, LGTM

PR #224, opened 2024-06-24
Updating Documentation for FetchContent Changes

FlorianXXIV:
After creating targets for CMakes FetchContent in #217 I updated the documentation so that the shown target is correct.
So 
from `target_link_libraries(${PROJECT_NAME} PRIVATE Graaf_lib)` 
to `target_link_libraries(${PROJECT_NAME} PRIVATE Graaf::Graaf)`

I also added a section about the CMake Options that are available, and gave examples on how to use them when people include this library for their Project.

I verified the documentation building locally.


bobluppes (approved):
Thanks a lot for improving the documentation here!
LGTM

PR #230, opened 2024-08-21
[clang-format] Change .clang-format file and apply it

goodicus:
A script file "./script/format_code.sh" is added for applying formatting rules via script also.


goodicus:
Sorry, wrong repo.

PR #249, opened 2024-11-19
fix: Potential race for file access in DotTest

joweich:
It appears like CI runs for `DotTest` fail non-deterministically. See recent runs of the coverage CI:

- [PR-245](https://github.com/bobluppes/graaf/actions/runs/11777353577/job/32801585412?pr=245)
- [PR-246](https://github.com/bobluppes/graaf/actions/runs/11899516112/job/33158300868?pr=246)
- [PR-247](https://github.com/bobluppes/graaf/actions/runs/11891803869/job/33133311432?pr=247)

While I can't reproduce this locally, my suspicion is that multiple tests use `"./test.dot"` simultaneously, which then causes race conditions or file access conflicts due to concurrent write/read attempts to the same file path.
This PR provides a fix by making the filenames unique (683ed984ad624dad0d87bbf8741fcc4a63a7bfdf).
f826ab66d8ea36e7c7702a4447db9a1622c54b89 increases cross-platform compatibility and should be unrelated to this issue.

### DIFF
--- a/test/graaflib/io/dot_test.cpp
+++ b/test/graaflib/io/dot_test.cpp
@@ -74,7 +74,7 @@ inline std::pair<vertex_id_t, vertex_id_t> make_sorted_pair(
 
 TEST(DotTest, EmptyUndirectedGraph) {
   // GIVEN
-  const std::filesystem::path path{"./test.dot"};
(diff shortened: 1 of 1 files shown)


bobluppes (approved):
Thank you for noticing this and addressing it!!

I am fairly confident that this was indeed a race condition. LGTM 🚀

PR #250, opened 2024-11-22
Added a hash combiner for performance improvement.

laurence-atomic:
In our testing at Atomic Industries, we discovered that the performance of graaf suffers on graphs with >50 million edges. 

We use a hash-combiner at Atomic to improve performance on hashing tuples. Implementing this in graaf resulted in substantial speed-ups in multiple test cases. Unfortunately I can not share internal performance data, but it should be possible to replicate the speed-up on large (>50 million edge) public datasets.


include/graaflib/types.h
```diff
@@ -9,14 +9,20 @@ namespace graaf {
 using vertex_id_t = std::size_t;
 using edge_id_t = std::pair<vertex_id_t, vertex_id_t>;
 
+template <class T>
+inline void hash_combine(std::size_t& seed, const T& v) {
+  std::hash<T> hasher;
+  seed ^= hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```
joweich, line 15: If I understand correctly, this is [hash_combine of boost v1.33](https://www.boost.org/doc/libs/1_86_0/libs/container_hash/doc/html/hash.html#notes_hash_combine).
A reference to the source as a comment would be good here. The magic number and the shifts otherwise seem a little arbitrary.

joweich (commented):
Benchmarks do indeed indicate a significant leap in performance:

bm_results

include/graaflib/types.h
laurence-atomic: Ah, it appears that is the source of the hash combine function. I was only familiar with the function based on seeing it used in multiple contexts in production. I hadn't tracked down the source.

It appears that like many magic constants, the value of 0x9e3779b9 was chosen arbitrarily as the fractional component of the golden ratio multiplied by 2^32. Perhaps on modern hardware, using the golden ratio multiplied by 2^64 would show improved performance.

Sources:

http://burtleburtle.net/bob/hash/doobs.html
https://softwareengineering.stackexchange.com/questions/63595/tea-algorithm-constant-0x9e3779b9-said-to-be-derived-from-golden-ratio-but-the/63599#63599

Perhaps it's better in that case to directly use the Boost hash combiner?

laurence-atomic: The bit shifts are probably chosen arbitrarily too.

laurence-atomic: I'll speak with my team at Atomic Industries about making a novel hash combiner function, so we don't accidentally end up plagiarizing another project. We have a few mathematicians on the team with a background in combinatorics.

bobluppes: +1 I agree that referencing Boost would be helpful here. Looking at the [Boost Software License](https://www.boost.org/LICENSE_1_0.txt) under which the [ContainerHash](https://github.com/boostorg/container_hash?tab=readme-ov-file) library is licensed this may also be a requirement. No need to push changes to the PR, I can add this before merging :) 

The alternative would be to add Boost as a dependency to Graaf, but I would like to avoid this as I would like Graaf to be a standalone Boost::Graph alternative.

include/graaflib/types.h
```diff
@@ -9,14 +9,20 @@ namespace graaf {
 using vertex_id_t = std::size_t;
 using edge_id_t = std::pair<vertex_id_t, vertex_id_t>;
 
+template <class T>
+inline void hash_combine(std::size_t& seed, const T& v) {
```
bobluppes, line 13: Adding the Boost license:

```suggestion

/**
 * Based on boost::hash_combine. Since Boost is licensed under the Boost
 * Software License, we include a copy of the license here.
 * TODO(b.luppes): consider using Boost's new hash_combine implementation
 *
 * Boost Software License - Version 1.0 - August 17th, 2003
 *
 * Permission is hereby granted, free of charge, to any person or organization
 * obtaining a copy of the software and accompanying documentation covered by
 * this license (the "Software") to use, reproduce, display, distribute,
 * execute, and transmit the Software, and to prepare derivative works of the
 * Software, and to permit third-parties to whom the Software is furnished to
 * do so, all subject to the following:
 *
 * The copyright notices in the Software and this entire statement, including
 * the above license grant, this restriction and the following disclaimer,
 * must be included in all copies of the Software, in whole or in part, and
 * all derivative works of the Software, unless such copies or derivative
 * works are solely in the form of machine-executable object code generated by
 * a source language processor.
 *
 * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 * FITNESS FOR A PARTICULAR PURPOSE, TITLE AND NON-INFRINGEMENT. IN NO EVENT
 * SHALL THE COPYRIGHT HOLDERS OR ANYONE DISTRIBUTING THE SOFTWARE BE LIABLE
 * FOR ANY DAMAGES OR OTHER LIABILITY, WHETHER IN CONTRACT, TORT OR OTHERWISE,
 * ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 * DEALINGS IN THE SOFTWARE.
 *
 */
template <class T>
inline void hash_combine(std::size_t& seed, const T& v) {
```

bobluppes (approved):
Thanks for opening this PR @laurence-atomic! And wow, that is some graph with 50M edges 😅 Glad you were able to find such a great performance win (and thanks @joweich for providing insights into our benchmarks)!

If I understand correctly, Boost switched to a slightly different implementation as of [Boost 1.81](https://www.boost.org/doc/libs/1_81_0/libs/container_hash/doc/html/hash.html#ref_hash_combine) which improves the distribution over the output domain. Source code [here](https://github.com/boostorg/container_hash/blob/ee5285bfa64843a11e29700298c83a37e3132fcd/include/boost/container_hash/detail/hash_mix.hpp#L67). I am wondering whether we could gain anything by using this newer implementation, but I would leave this as a potential follow up. 

LGTM! I just added the Boost license and am more than happy to include this optimization in the next release of Graaf 🚀

include/graaflib/types.h
bobluppes: Ah, apparently I cannot push changes nor can I accept code suggestions on your fork @laurence-atomic.
Could you please accept/add the license if it looks good to you?

laurence-atomic: I agree that using the latest version of the Boost hash function will probably improve performance even more. This update is fast enough for our use-case though. I'll accept the change request and push it back to you.

PR #301, opened 2026-01-11
feat: Add support for laying out graphs horizontally

mlyoung101:
For [my project](https://github.com/mattyoung101/slingshot), the graphs work much better when laid out horizontally, so I added this as a feature upstream. It just adds a flag to `io::to_dot` called `horizontal`, that defaults to false to preserve existing behaviour. Tests added too :)

### DIFF
--- a/include/graaflib/io/dot.h
+++ b/include/graaflib/io/dot.h
@@ -43,6 +43,8 @@ const auto default_edge_writer{
  * edge_id_t and a graph::edge_t and serialize it to a string. Default
  * implementations are provided for primitive numeric types.
  * @param path Path to the output dot file.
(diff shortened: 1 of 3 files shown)


mlyoung101:
Hey @bobluppes, if you have the chance do you reckon you could have a look at this? It's a pretty short diff.

bobluppes (approved):
Nice, thanks! Small, well-tested, backward compatible. Merging.

bobluppes:
One heads up for visibility: I'm mulling over narrowing the library's scope long-term (possibly moving serialization like to_dot out of core). Not acting on that now, just flagging so it's not a surprise later. Thanks again for the contribution!

PR #302, opened 2026-01-18
 improve get_neighbors performance

lili2012:
I use graaf a lot and change get_neighbors my executable run time drop from 700s to 60s. Hope you could accept it. Thanks!

### DIFF
--- a/include/graaflib/graph.h
+++ b/include/graaflib/graph.h
@@ -169,7 +169,7 @@ class graph {
    * @param  vertex_id The ID of the vertex
    * @return vertices_t - A list of neighboring vertices
    */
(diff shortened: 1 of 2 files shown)


mccakit:
I merged it to here
https://github.com/mccakit/neograaf

bobluppes (approved):
Thanks for this — great catch, and a very reasonable optimization! Verified locally: builds cleanly against current main, all 752 existing tests pass (including the dedicated GetNeighbors coverage for both directed and undirected graphs), and clang-format clean. The reference-lifetime tradeoff (the returned reference is tied to the graph's internal state until that vertex is removed) matches the existing pattern already used by get_vertex()/get_edge() in this class, so this is consistent with the rest of the API. Merging, thanks again for contributing!

PR #311, opened 2026-04-28
Fix external dependency build in CMake

kotaweav:
Fixes bug introduced in 2a4715f

This fix allows Graaf to be used as an external dependency using CMake's `FetchContent_Declare`. With `CMAKE_SOURCE_DIR`, CMake will look for `packaging/graaf.pc` in the project using Graaf, rather than Graaf itself. `PROJECT_SOURCE_DIR` is used over
`CMAKE_CURRENT_SOURCE_DIR` to maintain consistency within the `CMakeLists.txt`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,7 +43,7 @@ install(
     DIRECTORY "${PROJECT_SOURCE_DIR}/include/"
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 )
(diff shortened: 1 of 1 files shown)


bobluppes:
Thanks for this fix, and sorry for the slow response! Closing in favor of #312, which just got merged: same underlying bug, but that approach also avoids publishing a graaf.pc with incorrect metadata when Graaf is consumed as a subproject. Appreciate you tracking down the root cause and the write-up.

PR #312, opened 2026-04-29
Fix FetchContent following 2a4715ff46a25048e06078ea20afc325d59f987a

### LINKED ISSUE #310: [Bug] Non-pkgconf install
Recent changes broke the FetchContent:
https://github.com/bobluppes/graaf/pull/306#issuecomment-4152423709

FredericSabot:
Fix FetchContent by only installing pkg-config files when graaflib is the top project.

Closes #310

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,8 +43,10 @@ install(
     DIRECTORY "${PROJECT_SOURCE_DIR}/include/"
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
 )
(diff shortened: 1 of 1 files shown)


bobluppes (approved):
Thanks for this, and for flagging the overlap with #311 yourself! Verified locally (built + installed both #311 and #312 against a real FetchContent consumer project): #311 fixes the crash but the resulting graaf.pc mislabels the consumer's own install prefix as "graaf"; this approach avoids generating that incorrect metadata entirely, and `CMAKE_SOURCE_DIR STREQUAL PROJECT_SOURCE_DIR` is also the more portable choice given this repo's CMake 3.11 floor. Merging this one.

PR #386, opened 2026-09-09
fix: floyd_warshall_shortest_paths breaks after remove_vertex

### LINKED ISSUE #377: floyd_warshall_shortest_paths assumes contiguous vertex IDs, breaks after remove_vertex
## Summary

`floyd_warshall_shortest_paths` sizes its distance matrix by `vertex_count()` and indexes it directly with raw `vertex_id_t` values, assuming vertex IDs are contiguous `0..n-1`. The graph class does not guarantee this once any vertex has been removed.

## Location

[`include/graaflib/algorithm/shortest_path/floyd_warshall.tpp:15`](https://github.com/bobluppes/graaf/blob/main/include/graaflib/algorithm/shortest_path/floyd_warshall.tpp#L15)

## Details

Vertex IDs are assigned by an ever-increasing counter (`add_vertex` in `graph.tpp`), and `remove_vertex` does not compact or reassign remaining IDs. So `vertex_count()` (the *number* of vertices) and the *range* of valid vertex IDs can diverge as soon as a vertex is removed.

## Reproduction

1. Add 4 vertices (IDs `0`-`3`).
2. Remove vertex `1`.
3. `vertex_count() == 3`, so `n = 3` in `floyd_warshall_shortest_paths`, but surviving IDs are `{0, 2, 3}`.
4. The loop calls `graph.get_neighbors(1)` for a vertex that no longer exists → `adjacency_list_.at(1)` throws `std::out_of_range`.
5. Even where no throw occurs, indexing `shortest_paths[from][3]` is out-of-bounds for a 3-wide matrix → undefined behavior / potential heap corruption.

No test exercises `remove_vertex` before calling `floyd_warshall_shortest_paths`.

## Impact

Crash or undefined behavior from entirely ordinary API usage (remove a vertex, then run the algorithm) — no misuse required.

bobluppes:
## Summary
- `floyd_warshall_shortest_paths` sized its distance matrix by `vertex_count()` and indexed it directly with raw `vertex_id_t` values, assuming vertex IDs are contiguous `0..n-1`. `remove_vertex` does not compact or reassign remaining IDs, so this assumption breaks as soon as any vertex has been removed — leading to out-of-bounds writes into the distance matrix (undefined behavior) from entirely ordinary API usage.
- **This is a breaking API change** : the return type changes from `std::vector<std::vector<WEIGHT_T>>` to `std::unordered_map<vertex_id_t, std::unordered_map<vertex_id_t, WEIGHT_T>>`, keyed by real vertex IDs — matching the convention `dijkstra_shortest_paths` already uses elsewhere in the library, instead of introducing a second "compact index" coordinate space that callers would have to reason about.
- Internally, the `O(V³)` relaxation still runs against a dense, compact-indexed matrix (vertex IDs mapped to `0..n-1` for the algorithm's duration) for cache-friendly performance — this is converted to the vertex-ID-keyed result in a single `O(V²)` pass at the end, so correctness doesn't come at the cost of the dense-matrix performance Floyd-Warshall is chosen for.
- Added `[[nodiscard]]` to the declaration, matching `dijkstra_shortest_paths`.
- Updated the docs page and all existing tests for the new return type; no existing test's expected values changed, only the type they're expressed as.

## Test plan
- [x] Added `GraphWithNonContiguousVertexIdsAfterRemoval` regression test: adds 4 vertices, removes one, and confirms `floyd_warshall_shortest_paths` runs correctly against the remaining non-contiguous IDs.
- [x] Verified against the pre-fix code with a standalone repro compiled with `-fsanitize=address,undefined` and libc++ hardened assertions (`_LIBCPP_HARDENING_MODE_DEBUG`): the old code reliably traps on an out-of-bounds `vector[]` write for this exact scenario (`vector[] index out of bounds`); the fixed code runs cleanly under the same sanitizer settings.
- [x] Full test suite (`Graaf_test`, 771 tests) passes.

Fixes #377

### DIFF
--- a/docs/src/algorithms/shortest-path/floyd-warshall.md
+++ b/docs/src/algorithms/shortest-path/floyd-warshall.md
@@ -17,10 +17,10 @@ Calculates the shortest path between any two vertices.
 ```cpp
 template <typename V, typename E, graph_type T,
           typename WEIGHT_T = decltype(get_weight(std::declval<E>()))>
-std::vector<std::vector<WEIGHT_T>> floyd_warshall_shortest_paths(
-    const graph<V, E, T>& graph);
+std::unordered_map<vertex_id_t, std::unordered_map<vertex_id_t, WEIGHT_T>>
+floyd_warshall_shortest_paths(const graph<V, E, T>& graph);
 ```
 
 - **graph** The graph to extract the shortest path from.
-- **return** Returns a 2D vector of the shortest path. If a path doesn't exist between two vertices, mark it as
-  TYPE_MAX.
+- **return** Returns a map of maps, where `result.at(from).at(to)` is the shortest distance from vertex `from` to
+  vertex `to`. If no path exists between two vertices, there is no corresponding entry.
--- a/include/graaflib/algorithm/shortest_path/floyd_warshall.h
+++ b/include/graaflib/algorithm/shortest_path/floyd_warshall.h
@@ -3,7 +3,7 @@
 #include <graaflib/graph.h>
 #include <graaflib/types.h>
 
-#include <vector>
+#include <unordered_map>
 
 namespace graaf::algorithm {
 /**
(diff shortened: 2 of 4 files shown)


bobluppes:
At some point I would like to revisit this, but for now not introducing breaking API changes

PR #392, opened 2026-09-11
fix: greedy and Welsh-Powell coloring ignore incoming edges on directed graphs

### LINKED ISSUE #379: Greedy and Welsh-Powell graph coloring produce invalid colorings on directed graphs
## Summary

Both `greedy_graph_coloring` and `welsh_powell_coloring` only inspect `graph.get_neighbors(vertex)`, which for a `directed_graph` returns **outgoing** edges only. A vertex is colored without consulting its predecessors, so two vertices that share a common successor can end up with the same color — violating the library's own "no two adjacent vertices share a color" guarantee.

## Locations

- [`include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp:23`](https://github.com/bobluppes/graaf/blob/main/include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp#L23)
- [`include/graaflib/algorithm/coloring/welsh_powell.h:20`](https://github.com/bobluppes/graaf/blob/main/include/graaflib/algorithm/coloring/welsh_powell.h#L20) / [`welsh_powell.tpp:37`](https://github.com/bobluppes/graaf/blob/main/include/graaflib/algorithm/coloring/welsh_powell.tpp#L37)

## Details

`add_edge` for a `DIRECTED` graph only inserts into `adjacency_list_[lhs]`, never `adjacency_list_[rhs]` (`graph.tpp:187`). Neither coloring algorithm's template is constrained to `graph_type::UNDIRECTED` (unlike Kruskal, Prim, and Bron-Kerbosch, which are all explicitly pinned to `UNDIRECTED`), so both can be instantiated on a `directed_graph` and will silently produce an improper coloring.

## Reproduction

3 vertices `A, B, C` with directed edges `A -> C` and `B -> C`, no edge between `A` and `B`. Processing order `A, B, C`: `A` gets color `0` (its own outgoing neighbor list doesn't include `C`... wait, it's `C`'s incoming edges that are invisible), `B` gets color `0`, and `C` is colored using only its own (empty) outgoing neighbor list — so `C` can end up colored `0` as well, despite the edges `A -> C` and `B -> C` requiring `C` to differ from both.

`GreedyGraphColoringTest.DisconnectedComponents` is parameterized over `directed_graph` but never constructs a vertex with in-degree ≥ 2 from differently-colored predecessors, so it doesn't trigger the bug. `welsh_powell_test.cpp` only instantiates `minimal_undirected_graph_type`, so the directed case is entirely untested.

## Impact

Silent, incorrect output (not a crash) for any directed graph with a "converging" vertex — a very common shape.

bobluppes:
## Summary

Both `greedy_graph_coloring` and `welsh_powell_coloring` only inspected `graph.get_neighbors(vertex)`, which for a `directed_graph` returns **outgoing** edges only. A vertex was colored without consulting its predecessors, so two vertices with no edge between them but a shared successor could end up with the same color as that successor — violating the library's own "no two adjacent vertices share a color" guarantee.

Fixes #379.

## Fix

Both algorithms keep their existing generic `template <typename GRAPH>` signature — **no breaking API change**. For directed graphs, each now precomputes a reverse-adjacency (predecessor) map in O(V+E) up front, and folds those predecessor colors into the same color-selection logic already used for outgoing neighbors. Undirected graphs are unaffected (`get_neighbors` is already symmetric there, so the extra step is skipped via `graph.is_directed()`).

An alternative, more idiomatic-looking fix would have been to pin these algorithms to `graph_type::UNDIRECTED` (as Kruskal/Prim/Bron-Kerbosch already do) — but that would break any existing caller instantiating them on a `directed_graph`, which the issue asked to avoid.

## Is directed-graph coloring well-defined?

Researched this before implementing: there are two distinct, legitimate generalizations of vertex coloring to digraphs in the literature — (1) coloring the *underlying undirected graph* (adjacent = an edge in either direction), and (2) the *dichromatic number* (Neumann-Lara, 1982), which only forbids monochromatic directed cycles and can admit far fewer colors (e.g. 1 color for any DAG). These are genuinely different problems.

graaf's own docstrings/tests already commit to definition (1) ("no two adjacent vertices share the same color"), and the fix implements exactly that. Added a short clarifying note to both docstrings (`greedy_graph_coloring.h`, `welsh_powell.h`) making this explicit, since the ambiguity is exactly what let the original bug go unnoticed.

## Tests

- `greedy_coloring_test.cpp`: new `ConvergingEdgesColoredDifferently` case (both directed and undirected fixtures).
- `welsh_powell_test.cpp`: new `WelshPowellDirectedTest` suite (previously untested on directed graphs) with the same regression case.

Since `std::unordered_map` iteration order is unspecified, a naive single-triangle reproduction can pass by luck even against the buggy code (on this platform, Welsh-Powell's degree-based ordering and the greedy algorithm's hash-map ordering could both accidentally hide the bug for a single instance). Verified both new tests actually fail against the pre-fix code before restoring the fix, and confirmed the full 774-test suite passes with the fix applied.

## Test plan

- [x] `greedy_coloring_test.cpp` and `welsh_powell_test.cpp` pass with the fix
- [x] Verified both new regression tests fail against the pre-fix implementation
- [x] Full test suite (774 tests) passes
- [x] `clang-format --dry-run --Werror` clean on all changed files

### DIFF
--- a/include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
+++ b/include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
@@ -17,9 +17,10 @@ std::unordered_map<vertex_id_t, int> greedy_graph_coloring(const GRAPH& graph) {
   // Get the vertices from the graph
   const auto& vertices = graph.get_vertices();
 
-  // graph::get_neighbors() only reports outgoing edges. For a directed graph,
-  // two vertices sharing a common predecessor must still get different
-  // colors, so we additionally need to take incoming edges into account.
+  // graph::get_neighbors() only reports outgoing edges. On a directed graph,
+  // a vertex must also differ in color from each of its predecessors (the
+  // vertices with an edge into it), not just its successors, so we
+  // additionally need to take incoming edges into account.
   std::unordered_map<vertex_id_t, std::unordered_set<vertex_id_t>>
       predecessors{};
   if (graph.is_directed()) {
--- a/include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
+++ b/include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
@@ -2,6 +2,7 @@
 #include <graaflib/algorithm/coloring/greedy_graph_coloring.h>
 
 #include <unordered_map>
+#include <unordered_set>
 
 #include "greedy_graph_coloring.h"
 
@@ -15,17 +16,43 @@ std::unordered_map<vertex_id_t, int> greedy_graph_coloring(const GRAPH& graph) {
   // Get the vertices from the graph
   const auto& vertices = graph.get_vertices();
 
+  // graph::get_neighbors() only reports outgoing edges. For a directed graph,
+  // two vertices sharing a common predecessor must still get different
+  // colors, so we additionally need to take incoming edges into account. We
+  // precompute them here to keep the main loop below linear in the number of
+  // edges.
+  std::unordered_map<vertex_id_t, std::unordered_set<vertex_id_t>>
+      predecessors{};
+  if (graph.is_directed()) {
+    for (const auto& [vertex_id, _] : vertices) {
+      for (const auto neighbor_id : graph.get_neighbors(vertex_id)) {
+        predecessors[neighbor_id].insert(vertex_id);
+      }
+    }
+  }
+
   // Iterate through each vertex
   for (const auto& [current_vertex_id, _] : vertices) {
     // Iterate through neighboring vertices
     // Find the smallest available color for the current vertex
     int available_color{0};
-    for (const auto neighbor_id : graph.get_neighbors(current_vertex_id)) {
-      if (coloring.contains(neighbor_id)) {
-        const auto neighbor_color{coloring.at(neighbor_id)};
-        if (neighbor_color >= available_color) {
-          available_color = neighbor_color + 1;
+
+    const auto consider_neighbor{[&](vertex_id_t neighbor_id) {
+      if (const auto it{coloring.find(neighbor_id)}; it != coloring.end()) {
+        if (it->second >= available_color) {
+          available_color = it->second + 1;
         }
+      }
+    }};
+
+    for (const auto neighbor_id : graph.get_neighbors(current_vertex_id)) {
+      consider_neighbor(neighbor_id);
+    }
+
+    if (const auto it{predecessors.find(current_vertex_id)};
+        it != predecessors.end()) {
+      for (const auto predecessor_id : it->second) {
+        consider_neighbor(predecessor_id);
       }
     }
 
--- a/include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
+++ b/include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
@@ -1,5 +1,6 @@
 #pragma once
 #include <graaflib/algorithm/coloring/greedy_graph_coloring.h>
+#include <graaflib/algorithm/utils.h>
 
 #include <unordered_map>
 #include <unordered_set>
@@ -18,17 +19,11 @@ std::unordered_map<vertex_id_t, int> greedy_graph_coloring(const GRAPH& graph) {
 
   // graph::get_neighbors() only reports outgoing edges. For a directed graph,
   // two vertices sharing a common predecessor must still get different
-  // colors, so we additionally need to take incoming edges into account. We
-  // precompute them here to keep the main loop below linear in the number of
-  // edges.
+  // colors, so we additionally need to take incoming edges into account.
   std::unordered_map<vertex_id_t, std::unordered_set<vertex_id_t>>
       predecessors{};
   if (graph.is_directed()) {
-    for (const auto& [vertex_id, _] : vertices) {
-      for (const auto neighbor_id : graph.get_neighbors(vertex_id)) {
-        predecessors[neighbor_id].insert(vertex_id);
-      }
-    }
+    predecessors = get_predecessors(graph);
   }
 
   // Iterate through each vertex
--- a/include/graaflib/algorithm/coloring/welsh_powell.tpp
+++ b/include/graaflib/algorithm/coloring/welsh_powell.tpp
@@ -15,9 +15,10 @@ template <typename GRAPH>
 std::unordered_map<vertex_id_t, int> welsh_powell_coloring(const GRAPH& graph) {
   using degree_vertex_pair = std::pair<int, vertex_id_t>;
 
-  // graph::get_neighbors() only reports outgoing edges. For a directed graph,
-  // two vertices sharing a common predecessor must still get different
-  // colors, so we additionally need to take incoming edges into account.
+  // graph::get_neighbors() only reports outgoing edges. On a directed graph,
+  // a vertex must also differ in color from each of its predecessors (the
+  // vertices with an edge into it), not just its successors, so we
+  // additionally need to take incoming edges into account.
   std::unordered_map<vertex_id_t, std::unordered_set<vertex_id_t>>
       predecessors{};
   if (graph.is_directed()) {
--- a/include/graaflib/algorithm/coloring/welsh_powell.tpp
+++ b/include/graaflib/algorithm/coloring/welsh_powell.tpp
@@ -25,6 +25,20 @@ std::unordered_map<vertex_id_t, int> welsh_powell_coloring(const GRAPH& graph) {
 
   std::sort(degree_vertex_pairs.rbegin(), degree_vertex_pairs.rend());
 
+  // graph::get_neighbors() only reports outgoing edges. For a directed graph,
+  // two vertices sharing a common predecessor must still get different
+  // colors, so we additionally need to take incoming edges into account. We
+  // precompute them here to keep this linear in the number of edges.
+  std::unordered_map<vertex_id_t, std::unordered_set<vertex_id_t>>
+      predecessors{};
+  if (graph.is_directed()) {
+    for (const auto& [vertex_id, _] : graph.get_vertices()) {
+      for (const auto& neighbor_id : graph.get_neighbors(vertex_id)) {
+        predecessors[neighbor_id].insert(vertex_id);
+      }
+    }
+  }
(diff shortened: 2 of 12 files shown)


bobluppes:
Looked into the Codecov patch-coverage note (the checks themselves already pass, so this isn't blocking).

The single flagged line is the closing `}` of `get_predecessors` in [`utils.tpp`](include/graaflib/algorithm/utils.tpp#L36) — not a genuinely untested code path. I verified this by building `utils_test.cpp` standalone with Clang source-based coverage instrumentation: every statement in `get_predecessors`, for both the directed case (including a vertex with an empty neighbor set) and the undirected case, is hit by the existing tests.

This is the same gcov artifact already present, unaddressed, on the closing brace of `get_transposed_graph` in this same file (pre-existing on `main`, visible as Codecov's `base` miss before this PR) — a function that builds a container locally and returns it by value can get a separate epilogue/destructor basic-block count on the closing brace that doesn't always match the `return` statement's own count. No additional test can change that.

test/graaflib/algorithm/coloring/greedy_coloring_test.cpp
```diff
@@ -122,6 +122,47 @@ TYPED_TEST(GreedyGraphColoringTest, CompleteGraph) {
...
+// vertices with no edge between them but a shared successor must still be
+// colored differently from that successor. For a directed_graph, this means
+// the coloring may not rely solely on outgoing edges (A -> C, B -> C).
+//
+// The vertices in each converging triple are added in a different order
+// (sink added last, then sink added first) since graph::get_vertices()
+// returns an unordered_map whose iteration order is unspecified: whether the
+// bug is exposed for a given triple can depend on whether the sink happens to
+// be visited before or after its predecessors, so both orderings are covered
+// to make this test independent of that iteration order.
```
bobluppes: This comment is overly specific and tied to the implementation details of the graph class and the algorithm. You may mention that this is a regression test for a reported issue, but please do not expose implementation details in such comments.

test/graaflib/algorithm/coloring/greedy_coloring_test.cpp
```diff
@@ -122,6 +122,47 @@ TYPED_TEST(GreedyGraphColoringTest, CompleteGraph) {
...
+
+  // Triple 1: predecessors added before the sink.
+  const auto vertex_a1{graph.add_vertex(1)};
+  const auto vertex_b1{graph.add_vertex(2)};
+  const auto vertex_c1{graph.add_vertex(3)};
+  graph.add_edge(vertex_a1, vertex_c1, 1);
+  graph.add_edge(vertex_b1, vertex_c1, 1);
+
+  // Triple 2: sink added before its predecessors.
+  const auto vertex_c2{graph.add_vertex(4)};
```
bobluppes, line 143: Here you may briefly add a single line comment why you do before/after sync (i.e. to possibly trigger different iteration orders). But do not expose implementation details

test/graaflib/algorithm/coloring/welsh_powell_test.cpp
```diff
@@ -155,4 +155,41 @@ TYPED_TEST(WelshPowellTest, DisconnectedComponents) {
...
+// Regression test for https://github.com/bobluppes/graaf/issues/379:
+// graph::get_neighbors() only reports outgoing edges for a directed_graph, so
+// a vertex reached only via incoming edges must still be colored differently
+// from its predecessor.
+//
+// vertex_y has the highest degree (out-degree 2), so Welsh-Powell colors it
+// first regardless of unordered_map iteration order; vertex_z and vertex_x
+// are pure sinks (out-degree 0) whose only edge is incoming from vertex_y.
+// Once vertex_y is colored, a coloring that ignores incoming edges will color
+// both sinks the same as vertex_y, since it never looks back at it.
```
bobluppes: Same here, don't expose implementation details. You may mention briefly that this is a regression test for a specific issue.

test/graaflib/algorithm/coloring/welsh_powell_test.cpp
```diff
@@ -155,4 +155,41 @@ TYPED_TEST(WelshPowellTest, DisconnectedComponents) {
...
+TYPED_TEST(WelshPowellDirectedTest, ConvergingEdgesColoredDifferently) {
+  // GIVEN
+  using graph_t = typename TestFixture::graph_t;
+  graph_t graph{};
+
+  const auto vertex_y{graph.add_vertex(1)};
+  const auto vertex_z{graph.add_vertex(2)};
+  const auto vertex_x{graph.add_vertex(3)};
+
```
bobluppes: The order of these variable names is weird, I would expect x, y, z or otherwise a, b, c. Please adapt.

test/graaflib/algorithm/coloring/greedy_coloring_test.cpp
```diff
@@ -122,6 +122,47 @@ TYPED_TEST(GreedyGraphColoringTest, CompleteGraph) {
...
+  // WHEN
+  auto coloring = greedy_graph_coloring(graph);
+
+  // THEN
+  // Verify that for each edge, adjacent vertices have different colors
+  for (const auto& [edge_id, edge] : graph.get_edges()) {
+    const auto [u, v]{edge_id};
+    ASSERT_NE(coloring[u], coloring[v]);
+  }
+}
```
bobluppes, line 154: Welsh powell tests use a is_proper_coloring helper function which is defined in the anonymous namespace of that class. Please check: do we already have a place for shared utilities in the test sources (not in the main include sources!). 

If so, please extract the test utility there. If not, please leave as-is and open a github issue to create a test utility source and extract this coloring check helper in scope of that issue.

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
```diff
@@ -15,18 +17,38 @@ std::unordered_map<vertex_id_t, int> greedy_graph_coloring(const GRAPH& graph) {
   // Get the vertices from the graph
   const auto& vertices = graph.get_vertices();
 
+  // graph::get_neighbors() only reports outgoing edges. For a directed graph,
+  // two vertices sharing a common predecessor must still get different
```
bobluppes: This line reads strange to me. "Two vertices sharing a common predecessor must still get different color". 

I read this as:

a <- b -> c

Vertex a and c share a predecessor (b). According to the coment a and c must therefore get a different color. But my understanding is that the color between a & c does not matter, as long as they both have a different color compared to b.

Check if my understanding is correct. If so, please adjust the comment. If not, please explain where my understanding is not correct.

include/graaflib/algorithm/coloring/welsh_powell.tpp
```diff
@@ -16,10 +15,24 @@ template <typename GRAPH>
 std::unordered_map<vertex_id_t, int> welsh_powell_coloring(const GRAPH& graph) {
   using degree_vertex_pair = std::pair<int, vertex_id_t>;
 
-  // Step 1: Sort vertices by degree in descending order
+  // graph::get_neighbors() only reports outgoing edges. For a directed graph,
+  // two vertices sharing a common predecessor must still get different
+  // colors, so we additionally need to take incoming edges into account.
```
bobluppes: Same question here as for the comment in the greedy impleentation.

test/graaflib/algorithm/coloring/greedy_coloring_test.cpp
bobluppes: Trimmed the comment to just reference the issue, without the implementation-detail explanation.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

test/graaflib/algorithm/coloring/greedy_coloring_test.cpp
bobluppes: Kept a brief one-line note on why the sink is added before its predecessors here, without naming the underlying container/iteration mechanism.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

test/graaflib/algorithm/coloring/welsh_powell_test.cpp
bobluppes: Trimmed to just the issue reference here as well.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

test/graaflib/algorithm/coloring/welsh_powell_test.cpp
bobluppes: Renamed vertex_y/vertex_z/vertex_x to vertex_a/vertex_b/vertex_c (declaration order now matches the naming order), consistent with the a/b/c convention already used in greedy_coloring_test.cpp.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

test/graaflib/algorithm/coloring/greedy_coloring_test.cpp
bobluppes: Found the existing convention: test/utils/{fixtures,scenarios} already holds shared test-only utilities (not under include/). Extracted is_proper_coloring into test/utils/assertions/coloring_assertions.h and switched both welsh_powell_test.cpp (removing its anonymous-namespace copy) and this new greedy test (removing the manual edge-check loop) to use it.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
bobluppes: You're right, and the comment was wrong. For a <- b -> c, a and c don't need to differ from each other - only from b. The actual invariant is: a vertex must differ from each of its own predecessors, not that vertices sharing a predecessor must differ from each other. Reworded the comment to say that.

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

include/graaflib/algorithm/coloring/welsh_powell.tpp
bobluppes: Same fix applied here - reworded to state the correct invariant (a vertex must differ from each of its predecessors).

🤖 Addressed by [Claude Code](https://claude.com/claude-code)

PR #394, opened 2026-09-11
fix: remove_edge throws invalid_argument instead of unhandled out_of_range

### LINKED ISSUE #381: remove_edge throws unhandled std::out_of_range, inconsistent with the rest of the API
## Summary

`remove_edge` uses `adjacency_list_.at(...)`, which throws an undocumented, unhandled `std::out_of_range` whenever a vertex has no adjacency-list entry. This is inconsistent with the rest of the API, which throws `std::invalid_argument` with a descriptive message for invalid input.

## Location

[`include/graaflib/graph.tpp:204-220`](https://github.com/bobluppes/graaf/blob/main/include/graaflib/graph.tpp#L204-L220)

## Reproduction

```cpp
directed_graph<int, int> g;
auto a = g.add_vertex(1);
auto b = g.add_vertex(2);
g.add_edge(a, b, 10);
g.remove_edge(b, a);  // throws std::out_of_range
```

`b` was never used as an edge source, so `adjacency_list_` has no entry for it. `adjacency_list_.at(b).erase(...)` throws `std::out_of_range` instead of behaving like `has_edge(b, a) == false` (i.e. a no-op or a consistent `std::invalid_argument`). The same issue exists in the undirected branch if either vertex never participated in any edge.

## Impact

Reachable via ordinary misuse (swapped argument order, or removing an edge that doesn't exist) — not documented in `graph.h` (`remove_edge` has no `@throws`), and the exception type doesn't match the rest of the class's error-handling convention.

bobluppes:
## Summary

`remove_edge` used `adjacency_list_.at(...)`, which threw an undocumented, unhandled `std::out_of_range` whenever a vertex had no adjacency-list entry (e.g. it was never used as an edge source). This was inconsistent with the rest of the class's error-handling convention, which throws a descriptive `std::invalid_argument` for invalid input.

## Fix

`remove_edge` now checks `has_edge(...)` up front and throws `std::invalid_argument` with a descriptive message if no edge exists between the two vertices — mirroring the exact pattern already used by `get_edge` (same message format, same style).

This is a pure implementation/behavior fix, no API signature changes. Note it does change runtime behavior beyond just the crash case: removing a non-existent edge now always throws `invalid_argument` rather than sometimes silently no-op'ing (previously it would only silently no-op when both vertices already had adjacency-list entries, and crash otherwise). The linked issue explicitly names this as an acceptable resolution. No internal callers/algorithms depended on the old silent no-op.

Also updated the `remove_edge` doc comment in `graph.h` with the new `@throws`, and added a regression test (`RemoveNonExistentEdgeThrows`, covering both directed and undirected graphs) reproducing the exact scenario from the issue.

## Test plan

- [x] Added `GraphTest.RemoveNonExistentEdgeThrows` covering both directed and undirected graph types
- [x] Full test suite passes locally (773/773 tests)

Fixes #381

### DIFF
--- a/include/graaflib/graph.h
+++ b/include/graaflib/graph.h
@@ -210,6 +210,8 @@ class graph {
    *
    * @param  vertex_id_lhs The ID of the first vertex
    * @param  vertex_id_rhs The ID of the second vertex
+   * @throws invalid_argument exception - If no edge exists between the two
+   * vertices
    */
   void remove_edge(vertex_id_t vertex_id_lhs, vertex_id_t vertex_id_rhs);
 
--- a/include/graaflib/graph.tpp
+++ b/include/graaflib/graph.tpp
@@ -203,6 +203,13 @@ void graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::add_edge(vertex_id_t vertex_id_lhs,
 template <typename VERTEX_T, typename EDGE_T, graph_type GRAPH_TYPE_V>
 void graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::remove_edge(
     vertex_id_t vertex_id_lhs, vertex_id_t vertex_id_rhs) {
+  if (!has_edge(vertex_id_lhs, vertex_id_rhs)) {
+    // TODO(bluppes): replace with std::format once Clang supports it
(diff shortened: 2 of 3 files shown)


bobluppes:
Technically this is a breaking change due to changing an exception type, but LGTM

PR #395, opened 2026-09-11
fix: A* reports f-score instead of true path cost as total_weight

### LINKED ISSUE #378: A* reports f-score (g + heuristic) as total_weight instead of the true path cost
## Summary

`reconstruct_path` sets `graph_path::total_weight` from `vertex_info[end_vertex].dist_from_start`, but for A* that field actually stores the f-score (`g + heuristic`), not the true path cost `g`. `total_weight` is therefore wrong whenever the heuristic is non-zero at the target vertex.

## Location

- [`include/graaflib/algorithm/shortest_path/common.tpp:25`](https://github.com/bobluppes/graaf/blob/main/include/graaflib/algorithm/shortest_path/common.tpp#L25) (the bad read)
- [`include/graaflib/algorithm/shortest_path/a_star.tpp:82-87`](https://github.com/bobluppes/graaf/blob/main/include/graaflib/algorithm/shortest_path/a_star.tpp#L82-L87) (where `dist_from_start` is set to the f-score)

## Details

In `a_star.tpp`, `dist_from_start` is set to `tentative_g_score + heuristic(neighbor)` (the f-score), reusing a field that other shortest-path algorithms (Dijkstra, Bellman-Ford, BFS) use to store the true accumulated cost `g`. `reconstruct_path` is shared across all of these algorithms and unconditionally reports that field as `total_weight`.

## Reproduction

Using a heuristic that returns a nonzero constant at the goal (e.g. the repo's own `AStarHeuristicImpact` test, which uses `heuristic2` returning `10` for every vertex including the target): a path whose true cost is `3` gets `total_weight` reported as `13`.

The existing test only checks that one run's `total_weight` is *greater* than another's — it never asserts the absolute value, so the off-by-heuristic error passes silently.

## Impact

Any caller relying on `total_weight` as the actual path cost after an A* search gets a wrong answer whenever the heuristic doesn't vanish at the goal (i.e. essentially any non-trivial heuristic).

bobluppes:
## Summary

Fixes #378.

`a_star_search` reused `detail::path_vertex` (shared with Dijkstra/BFS/Bellman-Ford) to store the f-score (`g + heuristic`) in `dist_from_start`. `reconstruct_path` unconditionally reports that field as `graph_path::total_weight`, so `total_weight` was wrong whenever the heuristic was non-zero at the target vertex (e.g. a true cost of `3` was reported as `13` with a heuristic returning `10` at the goal).

## Fix

- Introduced a small A*-local `detail::a_star_open_set_item` type (`id` + `f_score`) used only to order the open-set priority queue.
- `vertex_info` (used for path reconstruction) now always stores the true accumulated cost (`g_score`) in `dist_from_start`, exactly like every other shortest-path algorithm.
- No public API changes — `path_vertex`, `reconstruct_path`, `graph_path`, and the `a_star_search` signature are untouched, so this is **not a breaking change**.

## Tests

- `AStarHeuristicImpact` previously only asserted a relative comparison (`ASSERT_LT`) between the two heuristics' `total_weight`, which let this bug pass silently (per the issue's own diagnosis). Updated it to assert the absolute expected cost (`3`) for both heuristics, directly covering the reported reproduction case.
- Full test suite (771 tests) passes locally, including all A* variants across weight types.

### DIFF
--- a/test/graaflib/algorithm/shortest_path/a_star_test.cpp
+++ b/test/graaflib/algorithm/shortest_path/a_star_test.cpp
@@ -237,11 +237,20 @@ TYPED_TEST(AStarShortestPathTest, AStarHeuristicImpact) {
       a_star_search(graph, start_vertex, target_vertex, heuristic2);
 
   // THEN
-  // Verify that the path with the underestimating heuristic is shorter
+  // The heuristic only affects the order in which vertices are explored, not
+  // the cost reported for the resulting path: total_weight must always be
+  // the true accumulated edge weight, never inflated by the heuristic value
+  // (e.g. heuristic2 returns 10 at every vertex, including the target, so a
+  // bug that reported the f-score instead of the true cost would surface
+  // here as 13 instead of 3).
   ASSERT_TRUE(path_with_underestimating_heuristic.has_value());
   ASSERT_TRUE(path_with_overestimating_heuristic.has_value());
-  ASSERT_LT(path_with_underestimating_heuristic->total_weight,
-            path_with_overestimating_heuristic->total_weight);
+
+  constexpr weight_t expected_total_weight{3};
+  ASSERT_EQ(path_with_underestimating_heuristic->total_weight,
+            expected_total_weight);
+  ASSERT_EQ(path_with_overestimating_heuristic->total_weight,
+            expected_total_weight);
 }
 
 }  // namespace graaf::algorithm
--- a/test/graaflib/algorithm/shortest_path/a_star_test.cpp
+++ b/test/graaflib/algorithm/shortest_path/a_star_test.cpp
@@ -152,8 +152,12 @@ TYPED_TEST(AStarShortestPathTest, AStarSuboptimalPath) {
   const auto path = a_star_search(graph, vertex_id_1, vertex_id_3, heuristic);
 
   // THEN
-  ASSERT_TRUE(path.has_value());  // Check if optional has a value
-  // Note: The path might not be the shortest, but it should still be valid
+  // The heuristic returns the same value for every vertex, so it is
+  // consistent (it trivially satisfies the triangle inequality) and does not
+  // break optimality: the shortest path (1 -> 3 directly, cost 4) is still
+  // found.
+  const graph_path<weight_t> expected_path{{vertex_id_1, vertex_id_3}, 4};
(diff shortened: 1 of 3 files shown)


test/graaflib/algorithm/shortest_path/a_star_test.cpp
```diff
@@ -237,11 +237,20 @@ TYPED_TEST(AStarShortestPathTest, AStarHeuristicImpact) {
       a_star_search(graph, start_vertex, target_vertex, heuristic2);
 
   // THEN
-  // Verify that the path with the underestimating heuristic is shorter
+  // The heuristic only affects the order in which vertices are explored, not
+  // the cost reported for the resulting path: total_weight must always be
+  // the true accumulated edge weight, never inflated by the heuristic value
+  // (e.g. heuristic2 returns 10 at every vertex, including the target, so a
+  // bug that reported the f-score instead of the true cost would surface
+  // here as 13 instead of 3).
```
bobluppes: Please remove the comment about the f-score bug here. It is weirdly specific and only relevant for the old code in which the bug was present. It holds no meaning now anymore and only causes confusion.

test/graaflib/algorithm/shortest_path/a_star_test.cpp
bobluppes, line 1: Please check if the other tests in the file are "underspecified" and not assert based on total path length. If so I think we should assert the expected path length in these tests.

test/graaflib/algorithm/shortest_path/a_star_test.cpp
bobluppes: Removed the comment about the old f-score bug — it only made sense against the pre-fix code. See commit bb4857f.

test/graaflib/algorithm/shortest_path/a_star_test.cpp
bobluppes: Checked the other tests in the file: AStarMinimalShortestPath, AStarSimpleShortestPath, and MultiplePathsTest already assert the full expected graph_path (vertices + total_weight). AStarSuboptimalPath was the one underspecified test — it only checked has_value(). Since its heuristic is a constant, it's consistent and doesn't affect optimality, so I tightened it to assert the exact expected path (vertices {1,3}, total_weight 4). See commit bb4857f.

PR #396, opened 2026-09-11
fix: vertex_degree/vertex_outdegree undercount self-loops in undirected graphs

### LINKED ISSUE #382: vertex_degree/vertex_outdegree undercount self-loops in undirected graphs
## Summary

`vertex_degree`/`vertex_outdegree` undercount self-loops in undirected graphs, because `add_edge` inserts the same vertex into its own adjacency set twice, which a `std::unordered_set` collapses into a single entry — so a self-loop contributes `1` to degree instead of the conventional `2`.

## Location

[`include/graaflib/properties/vertex_properties.tpp:17-19`](https://github.com/bobluppes/graaf/blob/main/include/graaflib/properties/vertex_properties.tpp#L17-L19) (combined with `include/graaflib/graph.tpp:191-197`)

## Reproduction

```cpp
undirected_graph<int, int> g;
auto v = g.add_vertex(1);
g.add_edge(v, v, 99);
vertex_degree(g, v);  // returns 1
```

`add_edge` calls `adjacency_list_[v].insert(v)` twice with the same value, so the neighbor set holds one entry, not two. `vertex_degree`/`vertex_outdegree` then report `1` for a self-loop, whereas the standard graph-theory convention (a self-loop contributes 2 to a vertex's degree) would expect `2`.

`test/graaflib/properties/vertex_properties_test.cpp` has zero self-loop test cases.

## Impact

Silent, wrong degree computation for any undirected graph containing self-loops — degree-based algorithms (e.g. some coloring heuristics, degeneracy calculations) would get a slightly wrong input.

bobluppes:
## Summary

`vertex_degree`/`vertex_outdegree` undercounted self-loops in undirected graphs. `add_edge` inserts the same vertex into its own adjacency set twice for a self-loop, which `std::unordered_set` collapses into a single entry — so a self-loop contributed `1` to degree instead of the conventional `2`.

I first checked whether the recent `add_edge`/`remove_edge` fixes (#393, #394) had already addressed this — they hadn't; those changes target duplicate-edge insertion and missing-edge removal, not self-loop degree accounting. I confirmed the bug still reproduced and fixed it.

## Fix

`vertex_outdegree` now adds 1 for undirected graphs when the neighbor set contains the vertex itself (i.e. it has a self-loop), since the set stores the self-loop only once but the standard convention counts it twice toward degree.

Directed graphs are unaffected: a self-loop there already contributes correctly via one out-edge and one in-edge.

## Test plan

- [x] Added regression tests for self-loop degree in both directed and undirected graphs (`test/graaflib/properties/vertex_properties_test.cpp`)
- [x] Full test suite passes (779 tests)

Fixes #382

### DIFF
--- a/test/graaflib/properties/vertex_properties_test.cpp
+++ b/test/graaflib/properties/vertex_properties_test.cpp
@@ -82,8 +82,8 @@ TEST(DirectedGraphPropertiesTest, VertexDegreeWithSelfLoop) {
   graph.add_edge(vertex_id_1, vertex_id_2, 200);
 
   // THEN
-  // In a directed graph, the self-loop already contributes one to the
-  // outdegree and one to the indegree of vertex_id_1.
+  // In a directed graph, the self-loop contributes one to the outdegree
+  // and one to the indegree of vertex_id_1.
   ASSERT_EQ(vertex_outdegree(graph, vertex_id_1), 2);
   ASSERT_EQ(vertex_indegree(graph, vertex_id_1), 1);
   ASSERT_EQ(vertex_degree(graph, vertex_id_1), 3);
@@ -123,9 +123,12 @@ TEST(UndirectedGraphPropertiesTest, VertexOutDegreeWithSelfLoop) {
 
   // THEN
   // The self-loop on vertex_id_1 contributes two to its degree, per the
-  // standard graph-theory convention.
+  // standard graph-theory convention. For an undirected graph, indegree
+  // is equal to outdegree.
   ASSERT_EQ(vertex_outdegree(graph, vertex_id_1), 3);
   ASSERT_EQ(vertex_outdegree(graph, vertex_id_2), 1);
+  ASSERT_EQ(vertex_indegree(graph, vertex_id_1), 3);
+  ASSERT_EQ(vertex_indegree(graph, vertex_id_2), 1);
 }
 
 TEST(UndirectedGraphPropertiesTest, VertexDegreeWithSelfLoop) {
--- a/test/graaflib/properties/vertex_properties_test.cpp
+++ b/test/graaflib/properties/vertex_properties_test.cpp
@@ -70,6 +70,25 @@ TEST(DirectedGraphPropertiesTest, VertexDegree) {
   ASSERT_EQ(vertex_degree(graph, vertex_id_4), 2);
 }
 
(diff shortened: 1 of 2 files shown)


test/graaflib/properties/vertex_properties_test.cpp
```diff
@@ -70,6 +70,25 @@ TEST(DirectedGraphPropertiesTest, VertexDegree) {
...
+  const auto vertex_id_2{graph.add_vertex(20)};
+
+  // WHEN
+  graph.add_edge(vertex_id_1, vertex_id_1, 100);
+  graph.add_edge(vertex_id_1, vertex_id_2, 200);
+
+  // THEN
+  // In a directed graph, the self-loop already contributes one to the
+  // outdegree and one to the indegree of vertex_id_1.
+  ASSERT_EQ(vertex_outdegree(graph, vertex_id_1), 2);
```
bobluppes, line 87: This comment reads weird due to the "already". This is closely coupled to the old state of the code and the current diff. Please remove "already" from this comment.

test/graaflib/properties/vertex_properties_test.cpp
```diff
@@ -91,6 +110,40 @@ TEST(UndirectedGraphPropertiesTest, VertexOutDegree) {
...
+  // THEN
+  // The self-loop on vertex_id_1 contributes two to its degree, per the
+  // standard graph-theory convention.
+  ASSERT_EQ(vertex_outdegree(graph, vertex_id_1), 3);
+  ASSERT_EQ(vertex_outdegree(graph, vertex_id_2), 1);
+}
+
+TEST(UndirectedGraphPropertiesTest, VertexDegreeWithSelfLoop) {
+  // GIVEN
+  undirected_graph<int, int> graph{};
```
bobluppes, line 136: Should we also test in-degree for such an undirected graph (currently we only cover degree and outdegree)

test/graaflib/properties/vertex_properties_test.cpp
bobluppes: Removed "already" from the comment.

test/graaflib/properties/vertex_properties_test.cpp
bobluppes: Added indegree assertions to the undirected self-loop test.

PR #400, opened 2026-09-12
feat: change prim_minimum_spanning_tree return type to tree<vertex_id_t, WEIGHT_T>

bobluppes:
## Summary
- `prim_minimum_spanning_tree` now returns `std::optional<tree<vertex_id_t, WEIGHT_T>>` instead of `std::optional<std::vector<edge_id_t>>`, rooted at `start_vertex`.
- The tree is built inline during Prim's existing traversal: since a candidate edge's `from` vertex is always already in the MST when the edge is accepted, we look up its `tree_node*` (tracked in a `vertex_id_t -> tree_node*` map, which also doubles as the MST membership check) and call `add_child` on it. No separate build pass is needed, and the algorithm's core logic (priority queue, lazy deletion of stale entries) is unchanged.
- Node identity is `vertex_id_t` and edge values are `WEIGHT_T`, not the graph's raw vertex/edge payload types — this keeps the result cheap to build and tied back to the source graph by id, consistent with `graph_path` and the rest of the algorithm return types in this library.
- This is intentionally scoped to Prim only. Kruskal will follow separately, since it needs a different construction strategy (Kruskal's union-find accepts edges in an order that doesn't correspond to any parent-first traversal, so it can't reuse `add_child` the same way).
- `tree`/`tree.h`/`tree.tpp` are unchanged from `main` — no new public API added there. See "Review history" below for how we got here.

## Review history
Earlier revisions of this PR added `tree`/`tree_node::operator==` to the production `tree` class so `prim_test.cpp` could assert `*mst == expected` directly. That went through three iterations (plain recursive → caught a real stack-overflow bug via `/code-review high --fix` on deep trees → rewritten iteratively → simplified to a flatten-sort-compare) before we stepped back: every version was a bad trade-off for a general-purpose type — recursion risks overflow on large trees, the iterative fix was hard to follow, and the simplified version silently required vertex values to be unique within a tree, a constraint `tree` has no way to enforce. It also had zero production callers; it existed purely for this test file.

Final approach: `tree`/`tree.tpp` are reverted to their pre-`operator==` state (identical to `main`), and the comparison lives as a small recursive helper local to `prim_test.cpp`'s anonymous namespace, implemented via `tree`'s existing public interface. Recursion there is fine — it only ever runs on the small, hand-built expected trees in this test file, so the stack-overflow concern doesn't apply, and we don't need to worry about duplicate vertex values since we control the test data.

Also fixed along the way: a redundant `in_mst` set in `prim.tpp` — `tree_node_for_vertex` already tracked the same membership, so keeping both in sync cost an extra hash insert per vertex for nothing.

## Test plan
- `prim_test.cpp` builds an expected `tree<vertex_id_t, int>` per scenario with the same `add_child` API and asserts via a local `trees_equal()` helper, instead of the original unordered edge-set comparison.
- `tree_test.cpp` is unchanged from `main`.
- [x] `Graaf_test` (785 tests) passes locally.
- [x] `Graaf_perf`'s `bm_prim` benchmark builds and runs unchanged (no source changes needed — `benchmark::DoNotOptimize` works fine with the move-only `optional<tree<...>>`).
- [x] Updated `docs/src/algorithms/minimum-spanning-tree/prim.md` to match the new signature and return value.

### DIFF
--- a/include/graaflib/tree.h
+++ b/include/graaflib/tree.h
@@ -22,6 +22,16 @@ class tree {
   [[nodiscard]] tree_node* root() { return root_.get(); }
   [[nodiscard]] const tree_node* root() const { return root_.get(); }
 
+  /**
+   * Two trees are considered equal if they have the same shape: every node
+   * has the same value, the same edge value towards its parent, and the same
+   * set of children (order-independent). Parent back-pointers are internal
+   * bookkeeping and are not part of a tree's identity, so they are ignored.
+   */
+  [[nodiscard]] bool operator==(const tree& other) const {
+    return *root_ == *other.root_;
+  }
+
   struct tree_node {
     // TODO(b.luppes): we are leaking implementation details regarding memory
     // management of children using std::unique_ptr. Consider providing a
@@ -31,6 +41,8 @@ class tree {
     std::vector<child_link> children{};
 
     [[nodiscard]] tree_node* add_child(EDGE_T edge_val, VERTEX_T child_val);
+
+    [[nodiscard]] bool operator==(const tree_node& other) const;
   };
 
   struct child_link {
--- a/test/graaflib/algorithm/minimum_spanning_tree/prim_test.cpp
+++ b/test/graaflib/algorithm/minimum_spanning_tree/prim_test.cpp
@@ -2,33 +2,70 @@
 #include <gtest/gtest.h>
 #include <utils/scenarios/scenarios.h>
 
-#include <unordered_set>
-#include <utility>
+#include <algorithm>
+#include <tuple>
+#include <vector>
 
 namespace graaf::algorithm {
 
 namespace {
 
-using edge_set_t = std::unordered_set<edge_id_t, edge_id_hash>;
+using mst_tree_t = tree<vertex_id_t, int>;
 
-[[nodiscard]] bool compare_mst(const std::vector<edge_id_t>& actual_mst,
-                               edge_set_t expected_edges) {
-  for (const auto& edge : actual_mst) {
-    if (expected_edges.contains(edge)) {
-      expected_edges.erase(edge);
-      continue;
-    } else if (const edge_id_t inverse_edge{edge.second, edge.first};
-               expected_edges.contains(inverse_edge)) {
-      // Since the graph is undirected, we also check the inverse of the edge
-      expected_edges.erase(inverse_edge);
-      continue;
-    }
+struct tree_edge {
+  vertex_id_t parent;
+  vertex_id_t child;
+  int weight;
 
-    // The expected edges did not contain the edge, nor the inverse edge
+  [[nodiscard]] bool operator==(const tree_edge&) const = default;
+};
+
+void collect_edges(const mst_tree_t::tree_node* node,
+                   std::vector<tree_edge>& out) {
+  for (const auto& link : node->children) {
+    out.push_back(tree_edge{node->value, link.child->value, link.value});
+
+    // Every child must point back to the node we reached it from - this is
+    // an invariant of the tree class itself, but it is exactly what an
+    // algorithm producing the tree is responsible for getting right.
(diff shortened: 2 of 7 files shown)


include/graaflib/tree.h
```diff
@@ -22,6 +22,16 @@ class tree {
   [[nodiscard]] tree_node* root() { return root_.get(); }
   [[nodiscard]] const tree_node* root() const { return root_.get(); }
 
+  /**
+   * Two trees are considered equal if they have the same shape: every node
+   * has the same value, the same edge value towards its parent, and the same
+   * set of children (order-independent). Parent back-pointers are internal
+   * bookkeeping and are not part of a tree's identity, so they are ignored.
```
bobluppes: Let's remove the part about parent back pointers, this is an implementation detail which should not be exposed

test/graaflib/algorithm/minimum_spanning_tree/prim_test.cpp
```diff
@@ -2,33 +2,51 @@
...
+          trees_equal(*lhs_link.child, *rhs.children[i].child)) {
+        rhs_matched[i] = true;
+        found_match = true;
+        break;
+      }
+    }
+    if (!found_match) {
+      return false;
+    }
+  }
```
bobluppes: This looks very complicated! Let's just throw the children in an unordered set and call equality on that.

test/graaflib/algorithm/minimum_spanning_tree/prim_test.cpp
```diff
@@ -83,8 +103,10 @@ TEST(PrimMstTest, SingleEdge) {
   // THEN
   ASSERT_TRUE(mst.has_value());
 
-  const edge_set_t expected_edges{{start_vertex, vertex_1}};
-  ASSERT_TRUE(compare_mst(mst.value(), expected_edges));
+  mst_tree_t expected{start_vertex};
+  static_cast<void>(expected.root()->add_child(100, vertex_1));
```
bobluppes: I don't like these static casts to void, but I see add_child is nodiscard. Let's assign the return to a [[maybe_unused]] variable. 

I would also prefer to keep expected tree in the GIVEN section above for each test. Let's exctract the construction of the expected trees to an IIFE:

```
const auto expected{[&]() {
  mst_tree_t res{start_vertex};

  //...

  return res;
}()};
```

test/graaflib/algorithm/minimum_spanning_tree/prim_test.cpp
bobluppes: Fixed in 7f25be7 - replaced the matched-index nested loop with an `unordered_set<vertex_id_t>` equality check for the set of child vertex ids, then a short follow-up pass that checks edge weights and recurses into the matching subtrees. Much shorter and easier to follow, thanks.

test/graaflib/algorithm/minimum_spanning_tree/prim_test.cpp
bobluppes: Fixed in 7f25be7 - swapped the `static_cast<void>` calls for `[[maybe_unused]] auto*` variables, and moved each test's expected-tree construction into the GIVEN section via an IIFE as you suggested.

test/graaflib/algorithm/minimum_spanning_tree/prim_test.cpp
bobluppes: Fair, my first fix still hand-matched with find_if/all_of after building the set - that missed your actual point. Reworked in d347ee9: each child is now wrapped in a small `child_key{edge_value, node}` whose `operator==` recurses back into `trees_equal` (hash is trivial - just the child vertex's value, since it only needs to agree with equality, not capture the whole subtree). Comparing a node's children is now literally `children_of(lhs) == children_of(rhs)` - `unordered_set::operator==` does the order-independent matching itself, nothing hand-rolled.

PR #407, opened 2026-09-15
perf: add dijkstra_shortest_paths benchmark on large web-graph datasets

bobluppes:
## Summary
- Prim's and Kruskal's MST benchmarks (`perf/graaflib/prim_benchmark.cpp` / `kruskal_benchmark.cpp`) already download real-world web-graph datasets (SNAP `web-Google`, `web-BerkStan`) and benchmark against them.
- `dijkstra_shortest_paths` (single-source-to-all shortest paths) is a similarly-shaped, near-linear (`O((V+E) log V)`) algorithm, so it makes sense to benchmark it against the same large real-world datasets rather than only small synthetic graphs.
- Unlike Prim (which has to bound the connected component it runs on to avoid a known super-linear fringe-rescan issue), `dijkstra_shortest_paths` runs directly against the full graph and completes in single-digit seconds, so no bounding is needed here.

## Benchmark results
```
--------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations
--------------------------------------------------------------------------------------
bm_dijkstra_shortest_paths/web_google   1.0556e+10 ns   1.0515e+10 ns            1
bm_dijkstra_shortest_paths/web_berkstan 9896115616 ns   9791640878 ns            1
```

## Test plan


bobluppes:
Found and fixed (in the latest commit) a real bug affecting this benchmark: \`bm_dijkstra_shortest_paths\`'s function-local \`static const auto graph{...}\` was only initialized on the first of the two \`BENCHMARK_CAPTURE\` calls, so \`web_berkstan\` silently reused the \`web_google\` graph whenever both ran in the same process (the default). See #413 for the full root-cause writeup and the fix, which is now applied here too by caching per-dataset inside \`utils::construct_graph_from_file()\`.

\`\`\`
----------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations
----------------------------------------------------------------------------------
bm_dijkstra_shortest_paths/web_google   1.0376e+10 ns   1.0312e+10 ns            1
bm_dijkstra_shortest_paths/web_berkstan 9232185256 ns   9212074383 ns            1
\`\`\`
(Close to the originally-reported numbers in this case, since both datasets happen to cost a similar order of magnitude at full scale - but web_berkstan was technically measuring the wrong graph before this fix.)

PR #408, opened 2026-09-15
perf: add dijkstra_shortest_path benchmark on large web-graph datasets

bobluppes:
## Summary
- Prim's and Kruskal's MST benchmarks already download real-world web-graph datasets (SNAP `web-Google`, `web-BerkStan`) and benchmark against them; this adds the same treatment for `dijkstra_shortest_path` (single start/end pair).
- The end vertex for each dataset is a real vertex id taken from the tail of that dataset's file, so the search has to actually traverse a meaningful part of the graph rather than terminating after one hop.
- `dijkstra_shortest_path` runs directly against the full graph and completes in single-digit seconds, so no bounding of the input graph is needed here (unlike Prim, which has to bound the connected component it runs on due to a known super-linear fringe-rescan issue).

## Benchmark results
```
---------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations
---------------------------------------------------------------------------------
bm_dijkstra_shortest_path/web_google   8259782398 ns   8228193962 ns            1
bm_dijkstra_shortest_path/web_berkstan 9080111897 ns   8924888291 ns            1
```

## Test plan


bobluppes:
Found and fixed (in the latest commit) two issues here:

1. Same dataset-sharing bug as #413: \`bm_dijkstra_shortest_path\`'s function-local \`static const auto graph{...}\` was only initialized on the first \`BENCHMARK_CAPTURE\` call, so \`web_berkstan\` silently ran against the \`web_google\` graph. Fixed by caching per-dataset inside \`utils::construct_graph_from_file()\`.
2. Fixing that exposed the hardcoded end_vertex ids (916427 / 685230) as an arbitrary, incomparable pair once measured against their *real* respective graphs: 685230 turned out to be a short, ~1s hop from the start vertex in web-BerkStan, while 916427 remained a genuinely far, ~8.6s target in web-Google. Replaced both with a computed farthest-vertex lookup (via \`dijkstra_shortest_paths\`, run once outside the timed loop) so each dataset gets a real worst-case target.

Updated (correct) benchmark results, now comparable across datasets:
\`\`\`
---------------------------------------------------------------------------------
Benchmark                                       Time             CPU   Iterations
---------------------------------------------------------------------------------
bm_dijkstra_shortest_path/web_google   8829827990 ns   8743531914 ns            1
bm_dijkstra_shortest_path/web_berkstan 8540785834 ns   8476075222 ns            1
\`\`\`

PR #409, opened 2026-09-15
perf: add greedy_graph_coloring benchmark on large web-graph datasets

bobluppes:
## Summary
- Prim's and Kruskal's MST benchmarks already download real-world web-graph datasets (SNAP `web-Google`, `web-BerkStan`) and benchmark against them; this adds the same treatment for `greedy_graph_coloring`.
- The algorithm is a single `O(V+E)` pass over the vertices, so it runs directly against the full graph in a few seconds without needing to bound the input.

## Benchmark results
```
--------------------------------------------------------------------------------
Benchmark                                      Time             CPU   Iterations
--------------------------------------------------------------------------------
bm_greedy_graph_coloring/web_google   2577994576 ns   2571084652 ns            1
bm_greedy_graph_coloring/web_berkstan 2559896465 ns   2539080882 ns            1
```

## Test plan


bobluppes:
Found and fixed (in the latest commit) a real bug affecting this benchmark: \`bm_greedy_graph_coloring\`'s function-local \`static const auto graph{...}\` was only initialized on the first of the two \`BENCHMARK_CAPTURE\` calls, so \`web_berkstan\` silently reused the \`web_google\` graph (this is why the original numbers looked suspiciously close: 2.56s vs 2.58s). See #413 for the full root-cause writeup and the fix, which is now applied here too.

\`\`\`
--------------------------------------------------------------------------------
Benchmark                                      Time             CPU   Iterations
--------------------------------------------------------------------------------
bm_greedy_graph_coloring/web_google   2548345301 ns   2540190011 ns            1
bm_greedy_graph_coloring/web_berkstan 1473252192 ns   1472198433 ns            1
\`\`\`

PR #410, opened 2026-09-15
perf: add welsh_powell_coloring benchmark on large web-graph datasets

bobluppes:
## Summary
- Prim's and Kruskal's MST benchmarks already download real-world web-graph datasets (SNAP `web-Google`, `web-BerkStan`) and benchmark against them; this adds the same treatment for `welsh_powell_coloring`.
- The algorithm is an `O(V log V + E)` degree-sort followed by a single greedy coloring pass, so it runs directly against the full graph in a few seconds without needing to bound the input.

## Benchmark results
```
--------------------------------------------------------------------------------
Benchmark                                      Time             CPU   Iterations
--------------------------------------------------------------------------------
bm_welsh_powell_coloring/web_google   4426745143 ns   4416855029 ns            1
bm_welsh_powell_coloring/web_berkstan 4413685611 ns   4377785358 ns            1
```

## Test plan


bobluppes:
Found and fixed (in the latest commit) a real bug affecting this benchmark: \`bm_welsh_powell_coloring\`'s function-local \`static const auto graph{...}\` was only initialized on the first of the two \`BENCHMARK_CAPTURE\` calls, so \`web_berkstan\` silently reused the \`web_google\` graph (this is why the original numbers looked suspiciously close: 4.41s vs 4.43s). See #413 for the full root-cause writeup and the fix, which is now applied here too.

\`\`\`
--------------------------------------------------------------------------------
Benchmark                                      Time             CPU   Iterations
--------------------------------------------------------------------------------
bm_welsh_powell_coloring/web_google   4684553152 ns   4625213497 ns            1
bm_welsh_powell_coloring/web_berkstan 2749251043 ns   2735322112 ns            1
\`\`\`

PR #411, opened 2026-09-15
perf: add breadth_first_search benchmark on large web-graph datasets

bobluppes:
## Summary
- Prim's and Kruskal's MST benchmarks already download real-world web-graph datasets (SNAP `web-Google`, `web-BerkStan`) and benchmark against them; this adds the same treatment for `breadth_first_traverse`.
- `breadth_first_traverse` only marks a vertex as seen when it is popped from the queue, not when it is pushed, so an *exhaustive* traversal of the giant connected component in these datasets (~96-98% of all vertices, see the comment in `prim_benchmark.cpp` for exact figures) re-enqueues and re-scans high-degree vertices many times before they're marked seen. This fires the edge callback tens of millions of times and the benchmark never finishes in a reasonable time when run unbounded (confirmed locally: it did not complete within 5 minutes).
- `prim_benchmark.cpp` already works around this same underlying issue by bounding the connected component Prim's algorithm runs on. This PR reuses that exact approach (same `MAX_SUBGRAPH_VERTICES` cap and subgraph-building helper) to bound the traversal itself.

## Benchmark results
```
bm_breadth_first_search/web_google      2627321 ns      2611539 ns          270 subgraph_vertices_used=2.4k
bm_breadth_first_search/web_berkstan    2743079 ns      2631579 ns          268 subgraph_vertices_used=2.4k
```

## Test plan
- [x] Confirmed the unbounded version times out (>5 min) before adding the subgraph cap


bobluppes:
Bumped the subgraph cap from 2,400 to 45,000 vertices so the benchmark runs closer to 1 second instead of ~2.6ms, giving more stable measurements:
```
bm_breadth_first_search/web_google    822964016 ns    822059714 ns            1 subgraph_vertices_used=45k
bm_breadth_first_search/web_berkstan  870561311 ns    857761288 ns            1 subgraph_vertices_used=45k
```

bobluppes:
Correction to my previous comment: the numbers posted there were affected by a real bug (fixed in the latest commit, see #413 for the full writeup) where \`bm_breadth_first_search\`'s function-local \`static const auto graph{...}\` was only initialized on the first of the two \`BENCHMARK_CAPTURE\` calls, so \`web_berkstan\` silently ran against the \`web_google\` graph whenever both ran together (~789ms, matching web_google's own number, instead of its true ~1.27s). Fixed by caching per-dataset inside \`utils::construct_graph_from_file()\`.

The 45k vertex cap itself was tuned correctly despite the bug (I used a standalone single-dataset-per-process tool for that, which isn't affected by it), so no retuning was needed. Final, correct numbers:
\`\`\`
bm_breadth_first_search/web_google    868143560 ns    855275466 ns            1 subgraph_vertices_used=45k
bm_breadth_first_search/web_berkstan 1268309476 ns   1260842437 ns            1 subgraph_vertices_used=45k
\`\`\`

PR #412, opened 2026-09-15
perf: add bfs_shortest_path benchmark on large web-graph datasets

bobluppes:
## Summary
- Prim's and Kruskal's MST benchmarks already download real-world web-graph datasets (SNAP `web-Google`, `web-BerkStan`) and benchmark against them; this adds the same treatment for `bfs_shortest_path`.
- `bfs_shortest_path` is built on `breadth_first_traverse`, which only marks a vertex as seen when it is popped from the queue, not when it is pushed. On the giant connected component in these datasets (~96-98% of all vertices) this can re-enqueue and re-scan high-degree vertices many times before they're marked seen - the same issue the companion `breadth_first_search` benchmark PR (#411) hits.
- So, as with that PR, this runs the search against a bounded subgraph (reusing `prim_benchmark.cpp`'s existing approach) rather than the full graph. The target vertex is picked from within that subgraph (the max vertex id encountered), so it's guaranteed reachable from the start vertex.

## Benchmark results
```
bm_bfs_shortest_path/web_google         2793867 ns      2792503 ns          246 subgraph_vertices_used=2.4k
bm_bfs_shortest_path/web_berkstan       2804318 ns      2801729 ns          249 subgraph_vertices_used=2.4k
```

## Test plan


bobluppes:
Widened the subgraph cap from 2,400 to 50,000 vertices, and changed the target-vertex selection from "highest vertex id in the subgraph" (arbitrary, and in practice let the search terminate almost immediately) to "the vertex an exhaustive BFS over the subgraph dequeues last" (a genuine near-worst-case target). Together this brings the benchmark close to 1 second instead of ~2.8ms:
```
--------------------------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------
bm_bfs_shortest_path/web_google    978350569 ns    967333775 ns            1 subgraph_vertices_used=50k
bm_bfs_shortest_path/web_berkstan  973285537 ns    973031445 ns            1 subgraph_vertices_used=50k
```

bobluppes:
Correction to my previous comment: found and fixed (latest commit) the same dataset-sharing bug described in #413 - \`bm_bfs_shortest_path\`'s function-local \`static const auto graph{...}\` was only initialized on the first \`BENCHMARK_CAPTURE\` call, so \`web_berkstan\` could silently run against the \`web_google\` graph when both ran together. Fixed by caching per-dataset inside \`utils::construct_graph_from_file()\`.

The 50k cap and worst-case target vertex selection from the previous commit were tuned with a standalone single-dataset-per-process tool, which isn't affected by this bug, so the numbers hold up. Final, correct results:
\`\`\`
--------------------------------------------------------------------------------------------
Benchmark                                  Time             CPU   Iterations UserCounters...
--------------------------------------------------------------------------------------------
bm_bfs_shortest_path/web_google    925933566 ns    924363377 ns            1 subgraph_vertices_used=50k
bm_bfs_shortest_path/web_berkstan  859563908 ns    859291449 ns            1 subgraph_vertices_used=50k
\`\`\`

PR #414, opened 2026-09-16
refactor: extract compute_connected_subgraph into a shared perf util

bobluppes:
## Summary
- `prim_benchmark.cpp`, `breadth_first_search_benchmark.cpp`, and `bfs_shortest_path_benchmark.cpp` each defined a byte-for-byte identical `compute_connected_subgraph()` helper, used to bound how much of the large downloaded web-graph datasets they run against.
- Extracted it into `perf/graaflib/utils/connected_subgraph.h`/`.cpp`, next to the existing `dataset_reader` utility, and updated all three benchmarks to call `utils::compute_connected_subgraph(...)` instead of defining their own copy.
- No behavioral change: each benchmark still passes its own `start_vertex`/`max_vertices`, and the resulting subgraphs (and thus timings) are identical.

## Test plan
- [x] `Graaf_perf` builds clean
- [x] Ran `bm_prim`, `bm_breadth_first_search`, `bm_bfs_shortest_path` against the real downloaded datasets - results land in the same ballpark as before the refactor:
```
bm_bfs_shortest_path/web_google       907795360 ns    907250033 ns            1 subgraph_vertices_used=50k
bm_bfs_shortest_path/web_berkstan     929585852 ns    922911968 ns            1 subgraph_vertices_used=50k
bm_breadth_first_search/web_google   1170313993 ns   1013592216 ns            1 subgraph_vertices_used=45k
bm_breadth_first_search/web_berkstan 1517863920 ns   1416422989 ns            1 subgraph_vertices_used=45k
bm_prim/web_google                    967205400 ns    967046998 ns            1 subgraph_vertices_used=330k
bm_prim/web_berkstan                 1123926200 ns   1123038830 ns            1 subgraph_vertices_used=190k
```


bobluppes:
Fixed a coupling issue pointed out in review: \`compute_connected_subgraph()\` hardcoded \`no_data{}\` for vertices and \`1\` for edge weights, which only worked because \`utils::construct_graph_from_file()\` happens to always build graphs that way. Updated it to read the actual values off the input graph via \`get_vertex()\`/\`get_edge()\` instead, so it stays correct regardless of what the caller's graph holds.

Re-ran all three affected benchmarks against the real datasets, results unchanged:
\`\`\`
bm_bfs_shortest_path/web_google      1025136612 ns   1020064094 ns            1 subgraph_vertices_used=50k
bm_bfs_shortest_path/web_berkstan    1085769394 ns   1085152223 ns            1 subgraph_vertices_used=50k
bm_breadth_first_search/web_google    873797074 ns    864216151 ns            1 subgraph_vertices_used=45k
bm_breadth_first_search/web_berkstan 1290988641 ns   1289682501 ns            1 subgraph_vertices_used=45k
bm_prim/web_google                   1187962998 ns   1178925361 ns            1 subgraph_vertices_used=330k
bm_prim/web_berkstan                 1210249444 ns   1209801219 ns            1 subgraph_vertices_used=190k
\`\`\`

Issue #416, opened 2026-09-16
Add unit tests for perf/common (shared benchmark dataset/subgraph utilities)


bobluppes:
## Context

`perf/common/` (added in #414 / the follow-up common-utils extraction) holds graph-library-agnostic, header-only utilities shared between the graaf benchmarks and (planned) Boost Graph comparison benchmarks:

- `perf/common/dataset.h` - `read_dataset_edges(dataset, callback)`, streams the deduplicated edge list for a SNAP dataset file.
- `perf/common/connected_subgraph.h` - `compute_connected_subgraph(dataset, start_vertex, max_vertices)`, BFS-bounds a dataset to a smaller connected subgraph.

Neither has unit test coverage today. This was noticed concretely during the extraction: re-implementing the BFS-bounding logic (previously duplicated across `prim_benchmark.cpp`, `breadth_first_search_benchmark.cpp`, `bfs_shortest_path_benchmark.cpp`) changed the exact resulting subgraph for a given `(dataset, start_vertex, max_vertices)` compared to the old per-benchmark implementations - because of two behavioral differences (different neighbor-iteration order, and mark-seen-at-enqueue vs graaf's own mark-seen-at-pop). This wasn't caught by anything automated; it was only found by manually re-running benchmarks and noticing the numbers had shifted well beyond noise.

## Why this is hard today

- `perf/` and `test/` are separate CMake targets/executables (`Graaf_perf` vs `Graaf_test`), and `test/CMakeLists.txt` doesn't currently include or build anything under `perf/`.
- `compute_connected_subgraph(dataset, ...)` currently calls `read_dataset_edges()` directly, which requires the real downloaded SNAP dataset file on disk (gated behind `GRAAF_DOWNLOAD_PERF_DATASETS`). That's a bad dependency for a unit test - it shouldn't need network access or a multi-hundred-MB download to verify BFS-bounding logic.

## Proposed approach

1. Decouple the BFS-bounding algorithm in `connected_subgraph.h` from the dataset file I/O: have the core algorithm take a generic edge source (e.g. any iterable of `(source, target)` pairs, or a small template parameter) rather than a `dataset` enum + internal `read_dataset_edges()` call. The real benchmarks would feed it `read_dataset_edges()`'s output; tests would feed it a small hardcoded synthetic graph.
[...]

PR #417, opened 2026-09-17
refactor: extract graph-agnostic dataset/subgraph core into perf/common

bobluppes:
## Summary

First step towards adding a standalone Boost Graph Library comparison benchmark (discussed separately): both that new executable and the existing `Graaf_perf` benchmarks need to build their graphs from the exact same dataset, so the genuinely reusable, graph-library-agnostic pieces of `perf/graaflib/utils/` are extracted into a new header-only `perf/common/` facility.

- `perf/common/dataset.h` - `read_dataset_edges(dataset, callback)` streams the deduplicated undirected edge list for a SNAP dataset file. No dependency on any graph library.
- `perf/common/connected_subgraph.h` - `compute_connected_subgraph(dataset, start_vertex, max_vertices)` BFS-bounds a dataset to a smaller connected vertex/edge set, returned as plain data (not a graph object), so every benchmark comparing libraries against each other builds its own graph from exactly the same bounded input.

`perf/graaflib/utils/dataset_reader.{h,cpp}` and `connected_subgraph.{h,cpp}` become thin adapters over `perf/common`, building the `graaf::graph_t` (copying real vertex/edge values) from the shared, library-agnostic structure.

## A note on behavior (not a silent regression)

This does **not** byte-for-byte reproduce the previous graaf-specific `connected_subgraph` implementation's exact output. That one's precise result depended on two graaf-internal implementation details: `std::unordered_set`-based neighbor storage (hash-bucket iteration order) and `breadth_first_traverse`'s mark-seen-on-pop quirk. Neither is a meaningful, deliberately-chosen property - and neither could be replicated by a genuinely shared, graph-library-agnostic utility anyway, since Boost's `adjacency_list` iterates neighbors in yet another order. The new implementation is deterministic (file-read order) and produces the *true* induced subgraph on the discovered vertex set (the old one could miss edges among a handful of "boundary" vertices that were discovered but never popped before the cap hit).

This changed the resulting subgraphs enough (denser at the same vertex count) that the "close to 1 second" bounds tuned for Prim, BFS traversal, and BFS shortest path needed to be recalibrated. All three were retuned and reverified against the real downloaded datasets (see below). `breadth_first_search` and `bfs_shortest_path` also switched from one shared cap to per-dataset caps (matching Prim's existing pattern), since the two datasets' local density now diverges enough that one constant can't land both near 1 second.

## Follow-up

Opened #416 to track adding unit test coverage for `perf/common` - it has none today, partly because `compute_connected_subgraph()` currently reads the dataset file directly rather than accepting a generic edge source, which makes it awkward to test without the real downloaded dataset. Not blocking this PR; noted so it isn't forgotten before the Boost benchmark work builds further on top of this.

## Benchmark results (reverified against real downloaded datasets)
```
bm_prim/web_google                    963290153 ns    958599007 ns  subgraph_vertices_used=85k
bm_prim/web_berkstan                 1035738545 ns   1034829062 ns  subgraph_vertices_used=80k
bm_breadth_first_search/web_google    869000725 ns    868688228 ns  subgraph_vertices_used=7.5k
bm_breadth_first_search/web_berkstan  892626217 ns    891187538 ns  subgraph_vertices_used=1050
bm_bfs_shortest_path/web_google       1032796448 ns  1031954873 ns  subgraph_vertices_used=20k
bm_bfs_shortest_path/web_berkstan      824777849 ns   802361714 ns  subgraph_vertices_used=1.7k
```

## Test plan
- [x] Full clean build (`Graaf_perf` + `Graaf_test`)
- [x] `ctest`: 795/795 tests pass (no test depends on `perf/`, but confirms nothing else broke)
- [x] Re-ran and retuned all three affected benchmarks against the real downloaded datasets (results above)


perf/graaflib/utils/connected_subgraph.h
```diff
@@ -4,12 +4,13 @@
...
-// real-world dataset they run against.
+// Builds the subgraph made up of the first max_vertices vertices discovered
+// by a breadth-first exploration of dataset_name's edge list, starting from
+// start_vertex. Vertex/edge values are copied from the corresponding
+// entries in `graph`, which must already hold the full dataset (e.g. as
+// returned by construct_graph_from_file(dataset_name)).
 [[nodiscard]] graph_t compute_connected_subgraph(
-    const graph_t& graph, graaf::vertex_id_t start_vertex,
-    std::size_t max_vertices);
+    const graph_t& graph, const dataset& dataset_name,
```
bobluppes: This feels awkward. One has to pass in both an input `graph` as well as the corresponding `dataset_name`. It is easy to pass the wrong dataset name for a given graph.

Let's change the interface of `compute_connected_subgraph` in perf/graaflib/utils to:

```
[[nodiscard]] graph_t construct_connected_subgraph(
    const dataset& dataset_name, graaf::vertex_id_t start_vertex, std::size_t max_vertices);
```

Internally it can then invoke dataset_reader to read the full input graph and afterwards constrain it.

perf/graaflib/utils/dataset_reader.cpp
```diff
@@ -1,80 +1,31 @@
...
+        }
 
-    // The source dataset is directed and may contain reciprocal edges
-    // (source->target and target->source), which map to the same edge in
-    // this undirected graph. Skip edges we've already added.
-    if (!graph.has_edge(source, target)) {
-      graph.add_edge(source, target, UNIT_WEIGHT);
-    }
-  }
+        graph.add_edge(source, target, 1);
```
bobluppes: Let's change the `1` here to `UNIT_WEIGHT` for readability.

perf/CMakeLists.txt
```diff
@@ -99,8 +103,13 @@ add_executable(
 # Include src such that we can #include <graaflib/*> in the sources
 target_include_directories(${PROJECT_NAME}_perf PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../include)
 
+# perf/common holds the graph-library-agnostic dataset/subgraph utilities
+# shared between this target and perf/boost's comparison benchmarks; this
```
bobluppes: Technically there is no perf/boost yet, let's remove it from the comment and keep it more generic.

bobluppes:
Addressed all three comments:

1. Dropped the separate \`graph\` parameter from \`compute_connected_subgraph\` - renamed to \`construct_connected_subgraph(dataset_name, start_vertex, max_vertices)\`, which now loads the full graph itself internally (matching \`construct_graph_from_file\`'s naming). No more risk of passing a graph that doesn't match the given dataset.
2. \`dataset_reader.cpp\` now uses the named \`UNIT_WEIGHT\` constant instead of a bare \`1\`.
3. Removed the premature \`perf/boost\` reference from the CMake comments, kept generic.

Rebuilt and reverified all three affected benchmarks against the real datasets - numbers unchanged (this was a pure interface cleanup, no behavior change), and all 795 unit tests still pass.

PR #419, opened 2026-09-17
perf: add opt-in Boost Graph Library comparison executable

bobluppes:
## Summary
- Adds a separate `Graaf_perf_boost` executable that benchmarks Boost Graph Library's `dijkstra_shortest_paths` on the same SNAP datasets (web-Google, web-BerkStan) as the existing `perf/graaflib/dijkstra_shortest_paths_benchmark.cpp`, built on top of the shared `perf/common/` dataset utilities so both libraries run against exactly the same edge lists.
- Gated behind a new `GRAAF_BUILD_BOOST_PERF` CMake option, default `OFF`. When off (the default), Boost is never fetched or configured and the executable target doesn't exist - confirmed by configuring with no extra flags and observing no Boost-related CMake output and no `Graaf_perf_boost` target.
- When enabled, Boost is fetched via `FetchContent` from the official, SHA256-pinned `boost-1.84.0.tar.gz` GitHub release asset - a single tarball with all of Boost's git submodules already checked out - rather than cloning the `boostorg/boost` git superproject directly (which is split across ~160 submodules). `BOOST_INCLUDE_LIBRARIES=graph` then limits the actual configure/build to Boost.Graph and its own dependencies.
- Documented the new option in `docs/src/quickstart/installation/installation.md` alongside the existing `GRAAF_DOWNLOAD_PERF_DATASETS`.

## Benchmark results
Ran locally with `-DGRAAF_BUILD_BOOST_PERF=ON -DGRAAF_DOWNLOAD_PERF_DATASETS=ON` against the real downloaded datasets:

| Dataset | Graaf `dijkstra_shortest_paths` | Boost `dijkstra_shortest_paths` |
|---|---|---|
| web-Google | 13.39 s | 1.33 s |
| web-BerkStan | 14.11 s | 0.69 s |

## Test plan
- [x] Configured with `-DGRAAF_BUILD_BOOST_PERF=ON`, confirmed Boost is fetched from the pinned tarball and `Boost::graph` resolves
- [x] Built `Graaf_perf_boost` successfully
- [x] Ran `Graaf_perf_boost` against the real downloaded datasets and confirmed correct results (table above)
- [x] Configured with the option left at its default (`OFF`) and confirmed Boost is never fetched/configured and `Graaf_perf_boost` is not built


perf/CMakeLists.txt
```diff
@@ -116,3 +116,52 @@ target_link_libraries(
...
+if(GRAAF_BUILD_BOOST_PERF)
+  # Fetched from the "boost-1.84.0.tar.gz" GitHub release asset - a single
+  # tarball with all of Boost's git submodules already checked out into it -
+  # rather than cloning the boostorg/boost git superproject directly, which
+  # would otherwise require a separate clone per submodule. This gives a
+  # reproducible, hash-verified, single-download fetch regardless of
+  # whether - or which version of - Boost happens to be installed on the
+  # machine. BOOST_INCLUDE_LIBRARIES limits the actual configure/build to
+  # Boost.Graph and its own dependencies rather than all of Boost.
+  set(BOOST_INCLUDE_LIBRARIES graph)
```
bobluppes, line 133: Let's cut down this comment dramatically. It contains a lot of details which are not relevant here. Dont' mention what you did not do, only mention what you did do.

bobluppes: Trimmed to just what we're doing and why (hash-pinned tarball + `BOOST_INCLUDE_LIBRARIES` scoping), dropped the alternatives-not-chosen detail. See 6de3a11.

PR #429, opened 2026-09-18
ci: merge benchmark tracking into the documentation workflow

bobluppes:
## Summary

- Merges `benchmarks.yml` into `documentation.yml` as three sequenced jobs (`changes` -> `benchmark` -> `documentation`), replacing two independent `push`-triggered workflows that could otherwise race: `needs:` only orders jobs *within* a single workflow, so there was no way to guarantee `documentation.yml` (which reads from `benchmark-data` to publish the chart, see below) waited for a concurrently-running `benchmarks.yml` to finish writing to it.
- `changes` uses `dorny/paths-filter` to detect whether the push touched `include/**`, `perf/**`, `CMakeLists.txt`, or this workflow file - mirroring the old top-level path filter, since that logic can no longer live at the trigger level once both jobs share one `on: push`.
- `benchmark` is unchanged in substance from the current `benchmarks.yml` (full dataset-backed suite, dataset caching, BigO/RMS stripping, push to `benchmark-data`) - it just now runs conditionally (`if: needs.changes.outputs.perf == 'true'`) instead of via a workflow-level path filter.
- `documentation` runs `needs: [changes, benchmark]` with `if: always()`, so it always deploys - whether `benchmark` ran, was skipped (this push didn't touch perf-relevant paths), or even failed. A benchmark regression or a flaky SNAP dataset download shouldn't hold the docs site hostage.
- `documentation` now also stitches the chart `benchmark-action/github-action-benchmark` generates on `benchmark-data` (`dev/bench/{index.html,data.js}`) into the published site under `/benchmarks/`, since GitHub Pages only supports one published source per repo and that's already the docs' `gh-pages` branch. The checkout of `benchmark-data` uses `continue-on-error: true` and the merge step checks the directory exists first, so a missing/deleted `benchmark-data` branch degrades to "publish docs without a chart" instead of breaking the deploy.
- Note on visibility: if `benchmark` fails, `documentation` (via `if: always()`) still runs and succeeds, but the *workflow run's* overall conclusion is still reported as failed - GitHub marks a run as failed if any job in it failed, regardless of what a later `if: always()` job does. So a benchmark failure stays clearly visible (red run in the Actions tab / commit status) without blocking the docs deploy.

## Test plan

- [x] Validated the workflow YAML parses correctly (`python3 -c "import yaml; yaml.safe_load(...)"`) and has the expected three jobs.
- [x] Simulated the chart-stitching shell logic locally against the real `benchmark-data` branch content (via `git worktree`): confirmed `docs/book/benchmarks/{index.html,data.js}` land in the right place.
- [x] Simulated the missing-branch case (no `benchmark-data` checkout present): confirmed it logs a message and exits cleanly rather than failing.
- [ ] Confirm the merged workflow runs end-to-end on the first push to `main` after merge (in particular that `documentation` correctly waits for `benchmark` and that Pages serves both `/` and `/benchmarks/`).


.github/workflows/documentation-and-benchmarks.yml
bobluppes, line 1: Please also update the name of the workflow file

bobluppes, line 1: This file contains a lot of very long new comments. Keep comments to a minimum and VERY concise. Should only be used to point out something non-obvious or give a BRIEF rationale if it is needed.

bobluppes: Renamed to `documentation-and-benchmarks.yml` in 3dd5226.

bobluppes: Trimmed every comment down to a single line each, kept only where non-obvious (e.g. the always()-still-shows-red note). See 3dd5226.

Issue #433, opened 2026-09-18
Dense vertex/adjacency storage in graph for faster lookups and traversal


bobluppes:
## Proposal
Store vertices and adjacency in contiguous vectors indexed by vertex id (or an internal dense slot index) instead of `unordered_map`s, giving a layout close to Boost's `vecS`/`vecS`.

## Why
Profiling the BFS benchmarks showed hashing and node-based containers dominate. Contiguous neighbor lists (tracked separately as a PR) already give large gains; dense vertex storage removes the remaining hash lookups (`get_neighbors`, `get_vertex`, `has_vertex`) and would let algorithms use vectors instead of hash sets/maps for per-vertex bookkeeping.

## Depends on
#432 (graph-owned vertex ids). With ids assigned by the graph they stay dense, so no id-map indirection or sparse-id hybrid is needed.

## Things to decide
- `get_vertices()` / `get_edges()` can no longer be `unordered_map`s; they would become ranges yielding `(id, vertex)` / `(edge_id, edge)` so structured-binding loops keep working. `.size()`, `.contains()`, `.at()` on them go away.
- `remove_vertex` policy: tombstones vs. id reuse (see #432).
- Benchmark before/after with the existing perf suite and the Boost comparison executable, one step at a time.

Issue #432, opened 2026-09-18
Remove add_vertex(vertex, id) so the graph owns vertex id assignment


bobluppes:
## Proposal
Remove the public `add_vertex(auto&& vertex, vertex_id_t id)` overload so the graph fully encapsulates vertex id generation.

## Why
- Guarantees ids are graph-assigned, so they stay (mostly) dense. This is the prerequisite for dense vertex storage (see the follow-up issue on that).
- Simplifies the id supplier: no more skipping over ids taken by explicit inserts.

## Where it is used today
- `include/graaflib/algorithm/utils.tpp`: the transposed graph copies vertices under their original ids (used by Kosaraju). Replace with a graph-level `transposed()` or an internal/friend copy hook that preserves ids.
- `perf/graaflib/utils/dataset_reader.cpp` and `perf/graaflib/utils/connected_subgraph.cpp`: insert vertices under dataset/original ids. These would keep an `external_id -> vertex_id_t` map instead.
- Tests: `test/graaflib/graph_test.cpp`, `test/graaflib/algorithm/strongly_connected_components/kosaraju_test.cpp`.

## Things to decide
- `remove_vertex` still leaves holes. Either keep tombstoned slots, or reuse ids and document that stale ids are invalid. Reusing ids makes old handles silently point at new vertices.
- Docs should show the external-id-map pattern for users importing data with their own ids.

This is a breaking change to the public API.

PR #434, opened 2026-09-18
perf: store neighbor lists as contiguous vectors

bobluppes:
## What
The graph's adjacency lists are now `std::vector<vertex_id_t>` instead of `std::unordered_set<vertex_id_t>`. `get_neighbors()` returns `const neighbors_t&` (the alias is renamed from `vertices_t`), and does a single hash lookup instead of two.

## Why
Profiling the BFS benchmarks showed node-based containers dominating. Iterating an `unordered_set` chases a heap node per neighbor; a vector is contiguous. Step A of the plan discussed for #433 (dense storage), independent of #432.

## Results (Release, full datasets, ms)
| Benchmark | before | after |
|---|---|---|
| breadth_first_search / google | 449 | 248 |
| breadth_first_search / berkstan | 320 | 82 |
| bfs_shortest_path / google | 545 | 376 |
| bfs_shortest_path / berkstan | 336 | 105 |
| greedy_graph_coloring / google | 543 | 186 |
| greedy_graph_coloring / berkstan | 276 | 51 |
| welsh_powell_coloring / google | 1232 | 644 |
| welsh_powell_coloring / berkstan | 656 | 333 |
| dijkstra_shortest_path / google | 1159 | 987 |
| dijkstra_shortest_path / berkstan | 1060 | 815 |
| prim / google | 85 | 64 |
| prim / berkstan | 86 | 61 |
| add_edge (10k) | 0.92 | 0.70 |

Kruskal and add_vertex are unchanged. The remaining BFS cost is mostly the algorithm's own `unordered_set` of seen vertices, which is a separate follow-up.

## Notes for reviewers
- **Breaking:** `get_neighbors()` no longer has `.contains()`, and `vertices_t` is now `neighbors_t`. Membership checks are linear (internal callers in `vertex_properties.tpp` updated). Neighbors are now in edge insertion order.
- `remove_edge`/`remove_vertex` are O(degree) on the neighbor lists. `add_edge` still rejects duplicates, so lists never hold duplicates; an undirected self-loop is stored once.
- Bron-Kerbosch converts neighbor lists to sets locally, since it needs set operations. Its benchmarks are unchanged.
- Added tests for insertion order and self-loop neighbor handling. All 797 tests pass.


include/graaflib/graph.h
```diff
@@ -171,12 +171,13 @@ class graph {
    * the vertices you'd find in Adj[vertex_id] in an adjacency-list
    * representation. It does not include predecessors (in-neighbors). For an
    * undirected graph, out- and in-neighbors coincide, so this returns all
-   * adjacent vertices.
+   * adjacent vertices. Neighbors are listed in the order their edges were
```
bobluppes: Let's keep the insertion order undocumented such that we can change it later if we so wanted.

include/graaflib/graph.tpp
```diff
@@ -192,13 +193,16 @@ void graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::add_edge(vertex_id_t vertex_id_lhs,
...
                    std::forward<decltype(edge)>(edge));
     return;
   } else if constexpr (GRAPH_TYPE_V == UNDIRECTED) {
-    adjacency_list_[vertex_id_lhs].insert(vertex_id_rhs);
-    adjacency_list_[vertex_id_rhs].insert(vertex_id_lhs);
+    adjacency_list_[vertex_id_lhs].push_back(vertex_id_rhs);
+    // A self-loop is a single neighbor entry
+    if (vertex_id_lhs != vertex_id_rhs) {
+      adjacency_list_[vertex_id_rhs].push_back(vertex_id_lhs);
+    }
```
bobluppes, line 205: Why?? Is this not a behaviour change compared to before?

test/graaflib/undirected_graph_test.cpp
```diff
@@ -44,20 +46,59 @@ TEST(UndirectedGraphTest, GetNeighbors) {
...
   const auto neighbors_vertex_3{graph.get_neighbors(vertex_id_3)};
   ASSERT_EQ(neighbors_vertex_3.size(), 1);
-  ASSERT_TRUE(neighbors_vertex_3.contains(vertex_id_1));
+  ASSERT_NE(std::ranges::find(neighbors_vertex_3, vertex_id_1),
+            neighbors_vertex_3.end());
+}
+
+TEST(UndirectedGraphTest, NeighborsAreListedInEdgeInsertionOrder) {
+  // GIVEN
+  undirected_graph<int, int> graph{};
```
bobluppes: Let's remove this test regarding insertion order. It is overly constraining and may hurt us if we want to change the implementation in the future such that this no longer holds.

include/graaflib/graph.tpp
bobluppes: Ah I think I see why this is now, otherwise it would show up twice if one is to get neighbors on it. 

Consider what is the best approach:
1. doing the "deduplication" on insert (as you do now) and have an adjacency map which is basically in an "asymmetric shape"
2. doing the "deduplication" on the edge retrieval.

include/graaflib/graph.h
bobluppes: Done: dropped the ordering sentence from the doc comment.

test/graaflib/undirected_graph_test.cpp
bobluppes: Done: removed the insertion order test (kept the self-loop one).

include/graaflib/graph.tpp
bobluppes: Not a behaviour change: with the `unordered_set` a self-loop was stored once, and this preserves that. I'd keep option 1 (dedupe on insert). Option 2 would need `get_neighbors()` to dedupe on retrieval, which cannot be done while returning a `const&` to the stored list: it would have to return by value and allocate on every call, undoing the traversal gains. Keeping the invariant that the list has no duplicates also means the only asymmetry is the self-loop case, which `vertex_outdegree` already accounts for by adding one for undirected self-loops. Happy to revisit if you'd prefer the symmetric layout.

Issue #435, opened 2026-09-18
Decide id-reuse vs. tombstoning policy for remove_vertex


bobluppes:
## Context

Split out from #432 (removing the `add_vertex(vertex, id)` overload). While implementing that issue it became clear `remove_vertex` currently leaves permanent id holes: `vertex_id_supplier_` never re-scans ids below its current value after a removal, so removed ids are never reused.

#432 called this out as prerequisite groundwork for dense vertex storage, but explicitly scoped it out of that change.

## Things to decide

- **Keep tombstoned/unused slots** — simple, no behavior change needed, but ids stay sparse over churn-heavy usage (works against the goal of dense vertex storage).
- **Reuse freed ids** — keeps ids dense, but makes old cached/stored vertex ids silently resolve to a different, newer vertex after a remove+add cycle. This is a correctness footgun for any caller holding onto a `vertex_id_t` across a `remove_vertex` call.

Whichever direction is chosen should be documented as part of the id-ownership contract on `graph`.

PR #436, opened 2026-09-18
refactor: remove add_vertex(vertex, id) so the graph owns id assignment

### LINKED ISSUE #432: Remove add_vertex(vertex, id) so the graph owns vertex id assignment

bobluppes:
## Summary
- Removes the public `add_vertex(vertex, id)` overload so vertex ids are always graph-assigned and stay dense.
- `get_transposed_graph()` now preserves vertex ids across the transpose via a private, friend-only hook (`add_vertex_with_id`) instead of the removed public overload — kept as a free function, consistent with every other algorithm in `algorithm/`, rather than becoming a `graph` member.
- Perf benchmark harness utilities that relied on caller-chosen ids (`dataset_reader.cpp`, `connected_subgraph.cpp`) now keep a local external-id → vertex-id map instead; `connected_subgraph.cpp` now mirrors the existing boost implementation and no longer depends on the pre-built dataset graph at all.
- Adds an "Importing data with existing IDs" doc section showing the external-id-map pattern for users migrating data with their own ids.

Closes #432

The `remove_vertex` id-hole/reuse question raised in #432 is intentionally out of scope here — tracked separately in #435.

## Test plan
- [x] `Graaf_test`: 796/796 tests pass
- [x] `Graaf_perf`: built and ran `bm_breadth_first_search` and `bm_prim` against the real web-Google/web-BerkStan datasets end-to-end to exercise the new id-remapping at runtime


include/graaflib/graph.h
```diff
@@ -11,6 +11,22 @@ namespace graaf {
...
+template <typename VERTEX_T, typename EDGE_T, graph_type GRAPH_TYPE_V>
+class graph;
+
+template <typename VERTEX_T, typename EDGE_T>
+using directed_graph = graph<VERTEX_T, EDGE_T, graph_type::DIRECTED>;
+
+template <typename VERTEX_T, typename EDGE_T>
+using undirected_graph = graph<VERTEX_T, EDGE_T, graph_type::UNDIRECTED>;
+
+// Forward declared so it can be granted friend access below: it needs to
```
bobluppes: Let's not couple the docu here to transpose. Instead mention this is for algorithms which require fine grained control over Id assignment

include/graaflib/graph.h
```diff
@@ -227,6 +233,15 @@ class graph {
   void remove_edge(vertex_id_t vertex_id_lhs, vertex_id_t vertex_id_rhs);
 
  private:
+  template <typename V, typename E>
+  friend directed_graph<V, E> get_transposed_graph(
+      const directed_graph<V, E>& graph);
+
+  // Only reachable by the befriended get_transposed_graph(): the public API
```
bobluppes: Also here, don't couple this general docu to the specific transpose

include/graaflib/graph.h
```diff
@@ -227,6 +233,15 @@ class graph {
...
+  // deliberately does not let callers pick a vertex's id, so that the graph
+  // stays free to assign them densely.
```
bobluppes: Also never mention densely packing in public docu. Mention something like performance optimization. The densely packing is an implementation details we want to keep free to change

perf/graaflib/utils/connected_subgraph.h
```diff
@@ -4,10 +4,21 @@
...
+  graaf::vertex_id_t start_vertex;
+};
+
 // Builds the subgraph made up of the first max_vertices vertices discovered
 // by a breadth-first exploration of dataset_name's edge list, starting from
 // start_vertex.
-[[nodiscard]] graph_t construct_connected_subgraph(
+//
+// Unlike utils::construct_graph_from_file(), the returned graph's vertex IDs
+// are not the original dataset vertex IDs: the graph assigns its own dense
```
bobluppes: Dont memstiom desnse

PR #437, opened 2026-09-18
refactor: reuse vertex ids freed by remove_vertex

### LINKED ISSUE #435: Decide id-reuse vs. tombstoning policy for remove_vertex

bobluppes:
## Summary
Resolves #435, the id-lifecycle decision flagged as a prerequisite for #433 (dense vertex/adjacency storage).

- `add_vertex()` now reuses ids freed by `remove_vertex()` (LIFO, via a small free-list) before growing `vertex_id_supplier_`.
- `remove_vertex()` only frees an id for reuse if the vertex actually existed, so calling it with an id that was never assigned can't leak an arbitrary large id into the free-list and get handed out later.
- Why this over the alternatives:
  - **Pure tombstoning (never reuse)** works fine for today's `unordered_map`-backed storage, but once #433 switches to a vector indexed directly by `vertex_id_t`, an unreused hole is no longer free — the vector has to stay sized to the highest id *ever* assigned, so any add/remove-churn workload grows memory unboundedly even while the live vertex count stays flat. That defeats the point of #433.
  - **Full compaction** (relabeling ids so they're always exactly `0..N-1`) was rejected: removing one vertex would silently renumber others, invalidating ids callers may still be holding.
  - Reuse via free-list bounds memory to the high-water mark of *concurrently* live vertices, at the cost of a documented caveat: a stale id kept around after removal can, after enough churn, get reassigned to a different vertex.
- Documented the id-stability contract on `add_vertex()`/`remove_vertex()` in `graph.h` and added a "Vertex ID lifecycle" section to the quickstart docs.
- Added `GraphTest.AddVertexReusesIdFreedByRemoveVertex` covering both the reuse behavior and the never-assigned-id guard.

## Test plan
- [x] `Graaf_test`: 798/798 tests pass (full suite + new test)
- [x] Verified no existing test relies on ids never being reused (`remove_vertex` is used in exactly one other test, `FloydWarshallTest.GraphWithNonContiguousVertexIdsAfterRemoval`, which never calls `add_vertex` afterwards so it's unaffected)


include/graaflib/graph.h
```diff
@@ -248,6 +255,11 @@ class graph {
   edge_id_to_edge_t edges_{};
 
   size_t vertex_id_supplier_{0};
+
+  // IDs freed by remove_vertex(), handed out again by add_vertex() before
+  // growing vertex_id_supplier_. Keeps ids bounded by the high-water mark of
+  // concurrently live vertices rather than the total ever created.
+  std::vector<vertex_id_t> free_vertex_ids_{};
```
bobluppes, line 260: Let's put this and it's comment above the vertex supplier member.

include/graaflib/graph.tpp
```diff
@@ -134,10 +134,17 @@ graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::get_neighbors(
...
 template <typename VERTEX_T, typename EDGE_T, graph_type GRAPH_TYPE_V>
 vertex_id_t graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::add_vertex(auto&& vertex) {
-  while (has_vertex(vertex_id_supplier_)) {
-    ++vertex_id_supplier_;
+  vertex_id_t vertex_id;
+  if (!free_vertex_ids_.empty()) {
+    vertex_id = free_vertex_ids_.back();
+    free_vertex_ids_.pop_back();
+  } else {
+    while (has_vertex(vertex_id_supplier_)) {
```
bobluppes, line 148: Is this while strictly necessary? Under which conditions do we increment mire than once?

bobluppes: Yes, still needed — `add_vertex_with_id()` (the private hook `get_transposed_graph()` uses) inserts directly and doesn't touch `free_vertex_ids_` or `vertex_id_supplier_` at all. So right after a transpose, the new graph's supplier is still 0 even though all N of its vertices already exist. The first subsequent `add_vertex()` call on it then has to scan past all N of them in one go, not just increment once. Added a comment explaining this plus a regression test (`UtilsTest.TransposedGraphAddVertexSkipsExistingIds`) covering exactly that scenario.

test/graaflib/algorithm/utils_test.cpp
```diff
@@ -64,6 +64,32 @@ TEST(UtilsTest, TransposePreservesIsolatedVertices) {
...
+  const auto vertex_id_3 = graph.add_vertex(3);
+
+  graph_t transposed_graph = get_transposed_graph(graph);
+
+  // WHEN adding a new vertex to the transposed graph
+  const auto new_vertex_id = transposed_graph.add_vertex(4);
+
+  // THEN the new vertex gets an id distinct from all three preserved ones,
+  // rather than colliding with one of them - a collision would make
+  // add_vertex()'s emplace() a silent no-op, leaving the vertex count at 3
```
bobluppes: Let's not add this implementation detail here. Keep it high level

PR #438, opened 2026-09-19
perf: store vertices in a dense vector indexed by id

bobluppes:
## Summary
First of two staged changes for #433 (adjacency storage - `adjacency_list_` - is a planned follow-up PR, benchmarked separately; edges stay in `edges_` unchanged, out of scope for #433 per its own "Why").

- `vertices_` moves from `unordered_map<vertex_id_t, VERTEX_T>` to `vector<optional<VERTEX_T>>`, indexed directly by id. Viable now that ids are always graph-owned and kept dense (#432, #435). `has_vertex()`/`get_vertex()` become a bounds check + direct index instead of a hash lookup.
- `vertex_count()` gets its own counter (`vertex_count_`), since `vertices_.size()` now counts tombstoned slots too.
- `get_vertices()` changes from returning a direct reference to the internal map to a lazy `std::ranges` view yielding the same `(id, value)` shape. Every real caller in the codebase only ever destructured `(id, value)` pairs, or (in one place) called `.size()` when `vertex_count()` already existed for that - so the old signature was just leaking the concrete storage type for no real benefit. This also means the accessor's return type won't have to change again the next time internal storage does.
- One caller (`greedy_graph_coloring.tpp`) stored the view in `const auto&` before iterating - fixed to iterate directly, since a filtering range can't be iterated through a const reference (`filter_view::begin()` isn't const). Documented as a usage note on `get_vertices()`.
- `vertex_properties.tpp` hardcoded `unordered_map<...>::value_type` as an explicit lambda parameter type - generalized to `auto`.
- `kosaraju.tpp`'s `get_vertices().size() == 0` switched to `graph.vertex_count() == 0`, which was already the correct API for that.

## Test fix
`GreedyGraphColoringTest.BasicGraphColoring` hardcoded the exact coloring produced by the old map's iteration order, rather than verifying properness like its sibling tests (`ConvergingEdgesColoredDifferently`) already do via `is_proper_coloring()`. The greedy algorithm's output was never guaranteed to be a *specific* coloring, only a *proper* one - fixed to match the established pattern instead of re-hardcoding new order-dependent numbers.

## Benchmarks (Release, `perf/graaflib/add_vertex_benchmark.cpp`, before vs. after, same machine)

|---|---|---|---|
| `bm_add_primitive_numeric_vertex/1000` | 114,689 ns | 25,186 ns | 4.6x |
| `bm_add_primitive_numeric_vertex/262144` | 19,999,622 ns | 5,555,523 ns | 3.6x |
| `bm_add_primitive_numeric_vertex/10000000` | 861,098,317 ns | 199,576,602 ns | 4.3x |
| `bm_add_user_defined_vertex/1000` | 179,819 ns | 119,639 ns | 1.5x |
| `bm_add_user_defined_vertex/10000000` | 1,414,313,709 ns | 988,971,626 ns | 1.4x |

(User-defined vertex type owns a `vector<double>`, so its own move/construction cost dominates more, hence the smaller relative gain - still a solid, consistent win.)

## Test plan
- [x] `Graaf_test`: 799/799 passing on both g++-12 and clang++-16 (Debug)
- [x] `Graaf_perf`: `add_vertex_benchmark` run in Release, before vs. after, see table above


include/graaflib/graph.h
```diff
@@ -71,13 +72,23 @@ class graph {
...
 
   /**
-   * @brief Get the internal vertices
+   * @brief Get a read-only view over the graph's vertices
    *
-   * @return const vertex_id_to_vertex_t& Map from vertex id to the user
-   * provided vertex.
+   * If you store the result in a variable rather than using it directly in a
+   * range-based for loop, store it as non-const (e.g. `auto`, not
+   * `const auto&`) - a filtering range can't be iterated through a const
```
bobluppes: This is not too nice, it is a read only view so as a user i would expect i can store it in Const auto. Lets explore some other options here.

include/graaflib/graph.h
```diff
@@ -71,13 +72,23 @@ class graph {
...
+   * reference.
+   *
+   * @return A range yielding a (vertex_id_t, const VERTEX_T&) pair for every
+   * vertex currently in the graph.
    */
-  [[nodiscard]] const vertex_id_to_vertex_t& get_vertices() const noexcept {
-    return vertices_;
+  [[nodiscard]] auto get_vertices() const noexcept {
+    return std::views::iota(vertex_id_t{0}, vertices_.size()) |
```
bobluppes: The implementation here belongs in the .tpp file.

include/graaflib/graph.tpp
```diff
@@ -182,14 +190,15 @@ void graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::remove_vertex(
   }
 
   adjacency_list_.erase(vertex_id);
```
bobluppes, line 259: Do we even need to do bookkeeping on the adjacency list if the vertex did not exist?

include/graaflib/graph.tpp
```diff
@@ -163,17 +231,24 @@ vertex_id_t graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::add_vertex_with_id(
...
 void graph<VERTEX_T, EDGE_T, GRAPH_TYPE_V>::remove_vertex(
     vertex_id_t vertex_id) {
-  // Only ids that were actually in use get freed for reuse - otherwise
-  // removing an id that was never assigned would leak it into
-  // free_vertex_ids_ and have a later add_vertex() hand it out.
-  const bool existed{has_vertex(vertex_id)};
+  // A vertex_id that was never assigned can't be anyone's neighbor, so
+  // there's nothing to clean up in adjacency_list_/edges_, and it must not
+  // be freed for reuse (that would leak an arbitrary id into
+  // free_vertex_ids_ for a later add_vertex() to hand out).
```
bobluppes: Don't add what would happen if you were to do this. Simply stating it should not happen is enough

PR #448, opened 2026-09-19
feat: add PageRank algorithm

### LINKED ISSUE #336: Add PageRank algorithm
## Summary

Graaf has no PageRank implementation, or any other iterative graph-ranking algorithm.

## Current state

A search across `include/graaflib/` for "pagerank"/"page_rank" returns no results, and there is no `algorithm/ranking/` (or similar) category. There is a related historical request, [#90 "[ALGO] Page Rank"](../issues/90), which was closed without being implemented (closed due to inactivity, not as "won't do").

## Why this matters

PageRank (and the broader family of iterative eigenvector-based ranking algorithms it represents) is one of the most widely recognized graph algorithms outside of pure computer science — used for web ranking, recommendation systems, citation analysis, and influence scoring in social/organizational networks. Its absence is a notable gap for anyone evaluating the library specifically for graph-analytics use cases (as opposed to pathfinding/traversal use cases, which are already well covered).

## Suggested resolution

- Implement PageRank (the standard damping-factor, iterative-until-convergence formulation) for directed graphs, under a new `include/graaflib/algorithm/ranking/` (or similarly named) category.
- Expose the damping factor, convergence tolerance, and max-iteration count as parameters.
- Add corresponding documentation under `docs/docs/algorithms/`, including complexity and convergence behavior.

## Acceptance criteria

- [ ] PageRank is implemented for directed graphs.
- [ ] Damping factor, convergence tolerance, and iteration limit are configurable.
- [ ] Documentation added, including complexity and convergence characteristics.
- [ ] Unit tests cover a known small graph with hand-verifiable expected rankings, dangling nodes (no outgoing edges), and disconnected components.

Vir-007:
Closes #336

## What

Adds `pagerank` for directed graphs under a new `include/graaflib/algorithm/ranking/` category, following the same header/`.tpp` split as the other algorithms.

```cpp
[[nodiscard]] std::optional<std::unordered_map<vertex_id_t, double>> pagerank(
    const graph<V, E, graph_type::DIRECTED>& graph,
    double damping_factor = 0.85, double tolerance = 1.0e-6,
    std::size_t max_iterations = 100);
```

The damping factor, convergence tolerance and iteration limit are all configurable, as the issue asks.

## How

Standard power iteration. Every vertex starts at `1 / |V|`, and each iteration pushes rank along outgoing edges and adds the `(1 - d) / |V|` teleport term. Iteration stops once the L1 change in rank between two iterations is at or below `tolerance`.

**Dangling vertices** (no outgoing edges) are treated as linking to every vertex, so their rank is redistributed uniformly each iteration. This keeps the ranks summing to one and matches the usual convention (networkx does the same).

**Error reporting** follows the split described in #403:
- Invalid parameters (damping factor outside `[0, 1]` or NaN, non-positive tolerance, zero iterations) throw `std::invalid_argument`, since those are misuse.
- Not converging within `max_iterations` returns `std::nullopt`, since the input is well-formed and there is simply no answer within the budget. With `d < 1` the L1 error shrinks by at least a factor `d` per iteration, so the defaults always converge (within ~90 iterations); `nullopt` in practice means `d = 1` on a periodic graph or a very small iteration limit.
- An empty graph returns an empty map.

Ranks are keyed by `vertex_id_t` in an `unordered_map`, so graphs with non-contiguous vertex ids (after `remove_vertex`) work. Edge weights are ignored.

Complexity is `O(k * (|V| + |E|))` time for `k` iterations and `O(|V|)` memory.

## Tests

`test/graaflib/algorithm/ranking/pagerank_test.cpp` covers:
- a small graph and a star graph checked against the exact solutions of the PageRank equations (hand-derived as fractions; also cross-checked against `networkx.pagerank` to 9 decimals)
- dangling vertices, including the star whose center has no outgoing edges
- disconnected components, both symmetric and asymmetric
- empty graph, single vertex, no edges, self-loop, non-contiguous vertex ids, `d = 0`
- non-convergence (`d = 1` oscillation, and an iteration limit of 1)
- all invalid parameter cases

## Docs

`docs/src/algorithms/ranking/pagerank.md` covers the algorithm, complexity and convergence behaviour, and it is added to `SUMMARY.md`, `README.md` and `llms.txt`. The issue mentions `docs/docs/algorithms/`, but the docs now live under `docs/src/`, so I followed the existing layout.


bobluppes (approved):
Thanks for for the strong contribution 🙏 
LGTM

I opened #450 to add the respective benchmark as a follow up. Happy to take it up on my end, but maybe you are also interested?

Issue #443, opened 2026-09-19
Replace hash-map vertex-id bookkeeping in algorithms with dense, vector-backed storage


bobluppes:
## Proposal
Replace the `std::unordered_map<vertex_id_t, ...>` / `std::unordered_set<vertex_id_t>` bookkeeping used throughout `algorithm/` (visited sets, distance/parent maps, union-find, color maps, ...) with dense, `vector`-backed storage indexed directly by `vertex_id_t`.

## Why
Vertex ids are now graph-owned and kept dense (#432, #435), which is exactly what makes direct vector indexing viable - the same insight that motivated moving the `graph` class's own vertex/adjacency storage off `unordered_map` (#433, #438, #442). That work stopped at the `graph` class boundary; every algorithm that consumes a graph still does its own id-keyed bookkeeping via hash maps/sets, which is now the dominant remaining cost.

Comparing against Boost.Graph on the same datasets/algorithms (`perf/graaflib` vs. the opt-in `perf/boost` target) shows gaps of 3.4x-20.7x depending on the algorithm, and this pattern shows up in essentially all of them:
- `breadth_first_search`: `std::unordered_set<vertex_id_t> seen_vertices`, one hash insert per edge visited (~5M for web-Google).
- `dijkstra_shortest_path(s)`, `prim`: `std::unordered_map<vertex_id_t, ...>` for distance/tree-membership tracking.
- `kruskal`'s union-find: `parent`/`rank` are `unordered_map<vertex_id_t, vertex_id_t>` - union-find is normally array-backed precisely because it's called once per edge and needs to stay close to O(1); a hash map defeats that.
- Also present in: coloring (`greedy_graph_coloring`, `welsh_powell`), SCC (`kosaraju`, `tarjan`), topological sort (both variants), cycle detection, `a_star`, `bellman_ford`, `floyd_warshall`, `bron_kerbosch`.

This is the single highest-leverage fix available: it's the one change that benefits literally every benchmarked algorithm, and it's lower-risk than touching edge storage (see #444) since it only affects algorithm-local working state, not the public `graph` API.

## Things to decide
- Algorithms need an upper bound on live vertex ids to size a dense vector correctly. `graph::vertex_count()` is a *live* count, not a bound on id values (removed ids can leave gaps below the current id supplier), so it's not sufficient on its own. Either expose something like an id-range bound on `graph`, or have algorithms grow their vectors lazily on first use of a given id.
[...]

PR #449, opened 2026-09-19
perf: replace BFS's unordered_set with a plain, densely-indexed vector<bool>

bobluppes:
## Summary
First of a staged series for #443.

Originally added a shared `vertex_id_set`/`vertex_id_map<T>` type to back algorithm-local vertex-id bookkeeping with dense vectors instead of hash maps/sets. Per review, that was more machinery than the actual problem needs right now: only one caller (BFS) exists so far, and a reusable abstraction isn't justified until there's a second and third real use to share it with. Also dropped the idea of exposing a `graph::vertex_id_upper_bound()`-style method to pre-size these vectors - that would leak `graph`'s internal id bookkeeping into its public API for a need `graph` itself has no reason to know about.

**Current approach**: `breadth_first_traverse()`'s `seen_vertices` is now a plain `std::vector<bool>`, grown lazily by a small helper local to that one function - no new files, no new types, no new public API anywhere. This works safely because every algorithm entry point takes the graph by `const&`, so it can't be mutated mid-algorithm (unlike `graph`'s own `add_vertex()`, see #446) - there's no incremental-growth hazard to guard against.

Remaining algorithms (Dijkstra, Prim, Kruskal's union-find, coloring, SCC, topological sort, cycle detection, A*, Bellman-Ford, Floyd-Warshall, Bron-Kerbosch) are left for follow-up PRs, each converted and benchmarked individually - if a shared pattern actually emerges across several of them, that's the point to consider factoring it out, not before.

## Benchmarks (Release, `perf/graaflib/breadth_first_search_benchmark.cpp`, real datasets)

| | before (`unordered_set`) | after (`vector<bool>`) | speedup |
|---|---|---|---|
| web_google | 3327.9 ms | 1326.1 ms | 2.51x |
| web_berkstan | 2879.7 ms | 1470.0 ms | 1.96x |

(Re-benchmarked after simplifying to confirm the dropped wrapper type wasn't contributing any of this win - it measured within noise of the plain-vector version.)

Also narrows the gap against the Boost.Graph comparison target (`perf/boost`, `GRAAF_BUILD_BOOST_PERF`):
- web_google: 4.95x -> 3.55x
- web_berkstan: 12.59x -> 8.65x

## Test plan
- [x] `Graaf_test`: 803/803 passing on both g++-12 and clang++-16 (Debug)
- [x] `Graaf_perf`: `breadth_first_search_benchmark` run in Release, before vs. after, see table above


include/graaflib/algorithm/graph_traversal/breadth_first_search.tpp
```diff
@@ -16,12 +16,27 @@ void breadth_first_traverse(
     const graph<V, E, T>& graph, vertex_id_t start_vertex,
     const EDGE_CALLBACK_T& edge_callback,
     const SEARCH_TERMINATION_STRATEGY_T& search_termination_strategy) {
-  std::unordered_set<vertex_id_t> seen_vertices{};
+  // Indexed directly by vertex_id_t rather than an unordered_set: ids are
+  // dense, and the graph is unmodified for the duration of the traversal
+  // (it's taken by const reference), so a plain vector that grows as new
+  // ids are discovered is a direct, allocation-light fit.
+  std::vector<bool> seen_vertices{};
+  const auto mark_seen{[&](vertex_id_t id) {
```
bobluppes: This function returns a book but this is not clear from the lambda name. A better name would be check_not_previously_marked. Probably not perfect still, but let's try to find a more descriptive name

PR #451, opened 2026-09-19
perf: back kruskal's union-find with vectors instead of unordered_map

bobluppes:
## Summary
Second of a staged series for #443 (BFS landed in #449).

`parent`/`rank` move from `unordered_map<vertex_id_t, vertex_id_t>` to `vector<vertex_id_t>`, indexed directly by id. `do_make_set()` is called once per vertex up front (iterating `graph.get_vertices()`, before any edge is processed), so by the time a `vertex_id_t` reaches `do_find_set()`/`do_merge_sets()` its slot already exists - lazy growth is only needed in `do_make_set()` itself, unlike BFS where discovery happens throughout the traversal.

Union-find is a textbook case for this kind of conversion: it's designed around O(α(n)) amortized find/union assuming array-backed storage with path compression - wrapping it in a hash map adds a hash + bucket-chain lookup to what's supposed to be close to a raw array access, on *every single edge processed* (Kruskal calls find/merge once per edge in the sorted edge list).

## Benchmarks (Release, `perf/graaflib/kruskal_benchmark.cpp`, real datasets)

|---|---|---|---|
| web_google | 6878.5 ms | 2252.3 ms | 3.05x |
| web_berkstan | 5356.7 ms | 2684.2 ms | 2.00x |

- web_google: 3.48x -> **1.14x**
- web_berkstan: 3.20x -> 1.60x

## Test plan
- [x] `Graaf_perf`: `kruskal_benchmark` run in Release, before vs. after, see table above


include/graaflib/algorithm/minimum_spanning_tree/kruskal.tpp
```diff
@@ -2,33 +2,39 @@
...
 
 #include "kruskal.h"
 
 namespace graaf::algorithm {
 
-// Disjoint Set Union to maintain sets of vertices
+// Disjoint Set Union to maintain sets of vertices.
+//
+// parent/rank are indexed directly by vertex_id_t rather than keyed in an
+// unordered_map: do_make_set() is called once per vertex up front, before
```
bobluppes: If you put up a comment only explain what you did do. Don't explain what you did not do.

bobluppes: Trimmed, thanks.

Issue #444, opened 2026-09-19
Investigate storing edge weights alongside adjacency to avoid a second lookup per traversed edge


bobluppes:
## Proposal
Investigate storing edge weights (or the full edge payload) alongside neighbor ids in `adjacency_list_`, so weighted traversal algorithms can retrieve an edge's weight without a separate lookup into `edges_`.

## Why
`edges_` is still `unordered_map<edge_id_t, EDGE_T, edge_id_hash>` - deliberately kept that way in #438/#442, since folding it into adjacency storage would trade `has_edge`/`get_edge(u, v)` from O(1) to O(degree), and this library's own benchmark datasets have real high-degree hub vertices that would be hurt by that. That decision was correct given the evidence available at the time (no algorithm-level profiling yet, only the `graph` class's own operations).

Comparing against Boost.Graph now provides that evidence: `dijkstra_shortest_path(s)`, `prim`, `a_star`, and `floyd_warshall` all call `get_weight(graph.get_edge(current, neighbor))` inside their neighbor loop - i.e. a hash lookup into `edges_` for every edge the traversal already found via the now-O(1) `get_neighbors()`. These are precisely the algorithms showing the largest gaps against Boost (Dijkstra 7-21x, Prim 5-7x), and their access pattern is "I already know this is a neighbor, give me its weight" - not "is there an edge between these two arbitrary vertices" (which is what `has_edge`/`get_edge(u, v)` actually needs O(1) for).

## Depends on
#443 (dense vertex-id bookkeeping in algorithms) - a smaller, lower-risk, broadly-applicable win that should land first. This issue is more architecturally invasive (touches `has_edge`, both `get_edge` overloads, `add_edge`, `remove_edge`, `get_edges()`) and deserves its own dedicated benchmarking pass, same as vertex/adjacency storage was split from each other in #433.

## Things to decide
- Whether to store `(neighbor_id, weight)` pairs directly in `adjacency_list_` (Boost `vecS`-style), which would mean `get_neighbors()`'s return type needs to change (it's used by ~10+ algorithm files purely for ids today) - or introduce a separate, parallel dense structure just for weights, indexed the same way as `adjacency_list_`, to avoid touching `get_neighbors()`'s contract.
[...]

PR #452, opened 2026-09-19
perf: back prim's tree-membership map with a vector instead of unordered_map

bobluppes:
## Summary
Third of a staged series for #443 (BFS landed in #449, Kruskal in #451).

`tree_node_for_vertex` moves from `unordered_map<vertex_id_t, tree_node_t*>` to `vector<tree_node_t*>`, indexed directly by id. Unlike Kruskal's union-find (every vertex known up front), Prim discovers vertices incrementally as the frontier grows, so this needs the same lazy-growth shape as BFS's `seen_vertices` - except the vector needs to distinguish "not yet in the tree" from "in the tree", so it's grown with `nullptr` as the not-present sentinel instead of `vector<bool>`.

A couple of details worth calling out:
- Membership checks (`in_tree()`) never grow the vector - an id past the current size is definitionally not in the tree yet, so a plain bounds check answers "not present" without allocating a slot for a vertex that's only being *queried*, not inserted. Only `add_to_tree()` grows.
- `vertices_in_tree` tracks the live count separately, since (unlike an `unordered_map`) a vector's `.size()` includes null-filled slots, not just live entries.

## Benchmarks (Release, `perf/graaflib/prim_benchmark.cpp`, real datasets)

|---|---|---|---|
| web_google | 410.2 ms | 296.0 ms | 1.39x |
| web_berkstan | 433.0 ms | 308.5 ms | 1.40x |

- web_google: 2.78x -> 2.01x
- web_berkstan: 2.60x -> 1.85x

Smaller win than Kruskal's, as expected: `prim_minimum_spanning_tree()` still pays for a separate `graph.get_edge()` hash lookup per candidate edge on top of this - that's #444's scope, not this one.

## Test plan
- [x] `Graaf_perf`: `prim_benchmark` run in Release, before vs. after, see table above


include/graaflib/algorithm/minimum_spanning_tree/prim.tpp
```diff
@@ -32,48 +32,66 @@ std::optional<tree<vertex_id_t, WEIGHT_T>> prim_minimum_spanning_tree(
 
   tree<vertex_id_t, WEIGHT_T> mst_tree{start_vertex};
 
-  // Doubles as the "is this vertex in the MST yet" membership check: a
-  // vertex has been added to the tree if and only if it has an entry here.
   using tree_node_t = typename tree<vertex_id_t, WEIGHT_T>::tree_node;
-  std::unordered_map<vertex_id_t, tree_node_t*> tree_node_for_vertex{
-      {start_vertex, mst_tree.root()}};
+
+  // Indexed directly by vertex_id_t rather than an unordered_map: doubles as
```
bobluppes: Also here, don't mention what you did not do (unordered_map)

bobluppes: Trimmed, thanks.

Issue #445, opened 2026-09-19
dijkstra_shortest_paths copies the entire path on every relaxation instead of backtracking once


bobluppes:
## Bug
`dijkstra_shortest_paths` (the all-destinations-from-source variant) rebuilds each reachable vertex's full path on every relaxation that improves its distance:

```cpp
shortest_paths[neighbor].total_weight = distance;
shortest_paths[neighbor].vertices = shortest_paths[current.id].vertices;
shortest_paths[neighbor].vertices.push_back(neighbor);
```

`graph_path::vertices` is a `std::list<vertex_id_t>`, so this copy is one heap allocation *per vertex already in the path*, repeated every time a shorter distance is found for a vertex (there's no decrease-key, so a vertex can be relaxed more than once).

## Why this is a bug, not a design tradeoff
The correct pattern already exists in this codebase and is used correctly by the single-target `dijkstra_shortest_path`: track only a `prev_id` per vertex during the main loop (O(1) per relaxation), then reconstruct the path with a single backward pass at the end via the existing `detail::reconstruct_path()` helper in `common.tpp`. `dijkstra_shortest_paths` just doesn't reuse it.

## Evidence
Comparing `perf/graaflib` against the opt-in Boost.Graph comparison target (`perf/boost`, `GRAAF_BUILD_BOOST_PERF`) on the same datasets:

| benchmark | graaf | boost | boost is |
|---|---|---|---|
| `dijkstra_shortest_paths`/web-berkstan | 17,722.6 ms | 855.0 ms | 20.7x faster |
| `dijkstra_shortest_paths`/web-google | 13,785.5 ms | 1,552.0 ms | 8.9x faster |
| `dijkstra_shortest_path`/web-berkstan | 12,805.3 ms | 744.8 ms | 17.2x faster |
| `dijkstra_shortest_path`/web-google | 9,857.2 ms | 1,377.3 ms | 7.2x faster |

`dijkstra_shortest_paths` is consistently worse than `dijkstra_shortest_path` against the same Boost baseline - the delta between the two isolates this issue specifically, on top of the shared causes tracked in #443 and #444.

## Fix
Track `path_vertex<WEIGHT_T>` (id, dist_from_start, prev_id) per vertex during the main loop instead of a full `graph_path`, and call `detail::reconstruct_path()` once per reachable vertex at the end - the same shape `dijkstra_shortest_path` already uses, just looped over every vertex that got a finite distance instead of stopping at one target.

PR #453, opened 2026-09-19
perf: back dijkstra_shortest_path's vertex_info with a vector

bobluppes:
## Summary
Fourth of a staged series for #443 (BFS in #449, Kruskal in #451, Prim in #452).

`vertex_info` moves from `unordered_map<vertex_id_t, path_vertex<WEIGHT_T>>` to `vector<optional<path_vertex<WEIGHT_T>>>`, indexed directly by id and grown lazily as vertices are discovered - the same shape used for Prim's `tree_node_for_vertex` (#452), since Dijkstra also discovers vertices incrementally rather than knowing them all up front like Kruskal's union-find (#451).

**Why path reconstruction is inlined instead of reusing `reconstruct_path()`**: that shared helper in `common.tpp` is typed against `unordered_map<vertex_id_t, path_vertex<WEIGHT_T>>` and is also used by `a_star_search()`. Backtracking the path is only a handful of lines, so duplicating it here avoids forcing `a_star`'s conversion into this same PR - keeping with one-algorithm-per-PR. Once `a_star` (and the `dijkstra_shortest_paths` fix from #445) are converted too, that's the natural point to reconsider whether `reconstruct_path()` should become generic over the vertex_info type - not before, with only one real converted caller.

## Benchmarks (Release, `perf/graaflib/dijkstra_shortest_path_benchmark.cpp`, real datasets)

|---|---|---|---|
| web_google | 9519.0 ms | 6037.4 ms | 1.58x |
| web_berkstan | 12397.9 ms | 9358.0 ms | 1.32x |

- web_google: 6.91x -> 4.38x
- web_berkstan: 16.65x -> 12.56x

The gap stays large regardless - expected, since this algorithm still pays for a separate `graph.get_edge()` hash lookup per edge (#444's scope) and uses a lazy-deletion `std::priority_queue` instead of a real decrease-key heap, neither of which is in scope for #443.

## Test plan
- [x] `Graaf_test`: 817/817 passing on both g++-12 and clang++-16 (Debug), including the negative-weight-throw tests
- [x] `Graaf_perf`: `dijkstra_shortest_path_benchmark` run in Release, before vs. after, see table above


bobluppes:
> Once a_star (and the dijkstra_shortest_paths fix from https://github.com/bobluppes/graaf/issues/445) are converted too, that's the natural point to reconsider whether reconstruct_path() should become generic over the vertex_info type - not before, with only one real converted caller.

Please open an issue for this.

include/graaflib/algorithm/shortest_path/dijkstra_shortest_path.tpp
```diff
@@ -15,10 +17,25 @@ std::optional<graph_path<WEIGHT_T>> dijkstra_shortest_path(
       std::priority_queue<weighted_path_item, std::vector<weighted_path_item>,
                           std::greater<>>;
   dijkstra_queue_t to_explore{};
-  std::unordered_map<vertex_id_t, weighted_path_item> vertex_info;
 
-  vertex_info[start_vertex] = {start_vertex, 0, start_vertex};
-  to_explore.push(vertex_info[start_vertex]);
+  // Indexed directly by vertex_id_t rather than an unordered_map, grown
```
bobluppes: Same here, don't explain what you did not do (unordered_map)

bobluppes: Trimmed, thanks.

bobluppes:
Filed https://github.com/bobluppes/graaf/issues/463 to track converting a_star and revisiting reconstruct_path()'s signature at that point.

PR #454, opened 2026-09-19
perf: fix dijkstra_shortest_paths' eager path copy and densify its storage

### LINKED ISSUE #445: dijkstra_shortest_paths copies the entire path on every relaxation instead of backtracking once

bobluppes:
## Summary
Fifth of a staged series for #443 (BFS in #449, Kruskal in #451, Prim in #452, `dijkstra_shortest_path` in #453). Also fixes #445.

These two issues are tightly coupled in this one function: fixing the eager-path-copy bug means changing what's tracked per vertex from a full `graph_path<WEIGHT_T>` down to just `(dist_from_start, prev_id)` - which is exactly the shape the #443 conversion needs anyway. Doing them as two separate PRs would mean an awkward intermediate state (a vector wrapping the same bug), so they're combined here.

`vertex_info` is now indexed directly by `vertex_id_t` (a vector, grown lazily) instead of `unordered_map<vertex_id_t, graph_path<WEIGHT_T>>`, and tracks only `dist_from_start`/`prev_id` during the search - the same shape used in #452/#453. Each reachable vertex's full path (a `std::list<vertex_id_t>`) is now reconstructed once at the end via backtracking, instead of being copied wholesale on every relaxation that improves a vertex's distance (#445's bug).

The function's return type is a documented public contract (`unordered_map<vertex_id_t, graph_path<WEIGHT_T>>`, tests depend on it), so that's unchanged - it's built once, after the main loop, from the dense `vertex_info`.

## Benchmarks (Release, `perf/graaflib/dijkstra_shortest_paths_benchmark.cpp`, real datasets)

|---|---|---|---|
| web_google | 11991.7 ms | 10775.5 ms | 1.11x |
| web_berkstan | 15906.9 ms | 14541.4 ms | 1.09x |

- web_google: 7.73x -> 6.94x
- web_berkstan: 18.61x -> 17.01x

**Honest note**: this is a smaller win than I expected given #445's bug is now fixed. That suggests relaxation-repeat rates and path lengths on these real-world graphs are more modest than the theoretical worst case, and the dominant remaining cost is the same per-edge `graph.get_edge()` hash lookup and lazy-deletion priority queue already called out as #444's territory in #453 - not more #443-shaped work in this function.

## Test plan
- [x] `Graaf_test`: 817/817 passing on both g++-12 and clang++-16 (Debug), including `dijkstra_shortest_paths_test.cpp` which asserts actual path contents/weights
- [x] `Graaf_perf`: `dijkstra_shortest_paths_benchmark` run in Release, before vs. after, see table above


include/graaflib/algorithm/shortest_path/dijkstra_shortest_paths.tpp
```diff
@@ -10,24 +12,36 @@ template <typename V, typename E, graph_type T, typename WEIGHT_T>
...
   using weighted_path_item = detail::path_vertex<WEIGHT_T>;
   using dijkstra_queue_t =
       std::priority_queue<weighted_path_item, std::vector<weighted_path_item>,
                           std::greater<>>;
   dijkstra_queue_t to_explore{};
 
-  shortest_paths[source_vertex].total_weight = 0;
-  shortest_paths[source_vertex].vertices.push_back(source_vertex);
-  to_explore.push(weighted_path_item{source_vertex, 0});
+  // Indexed directly by vertex_id_t rather than an unordered_map, grown
```
bobluppes: Also here, don't mention what you did not do (unordered_map)

include/graaflib/algorithm/shortest_path/dijkstra_shortest_paths.tpp
```diff
@@ -10,24 +12,36 @@ template <typename V, typename E, graph_type T, typename WEIGHT_T>
...
+  // lazily as vertices are discovered. Tracks only (dist_from_start,
+  // prev_id) per vertex during the search - each reachable vertex's full
+  // path is reconstructed once at the end via backtracking, rather than
+  // copied on every relaxation.
```
bobluppes: Also here, don't mention what you did not do (copy on every relaxation)

include/graaflib/algorithm/shortest_path/dijkstra_shortest_paths.tpp
```diff
@@ -10,24 +12,36 @@ template <typename V, typename E, graph_type T, typename WEIGHT_T>
...
+  std::vector<std::optional<weighted_path_item>> vertex_info{};
+  const auto has_info{[&](vertex_id_t id) {
+    return id < vertex_info.size() && vertex_info[id].has_value();
+  }};
+  const auto set_info{[&](vertex_id_t id, weighted_path_item item) {
+    if (id >= vertex_info.size()) {
+      vertex_info.resize(id + 1);
+    }
+    vertex_info[id] = item;
+  }};
```
bobluppes, line 34: In general we now have this vector + has_x + set_x pattern spread widely through the codebase. No action item for now, but please open an issue to investigate if we can encapsulate this and extract it to a common utility.

include/graaflib/algorithm/shortest_path/dijkstra_shortest_paths.tpp
```diff
@@ -10,24 +12,36 @@ template <typename V, typename E, graph_type T, typename WEIGHT_T>
 [[nodiscard]] std::unordered_map<vertex_id_t, graph_path<WEIGHT_T>>
 dijkstra_shortest_paths(const graph<V, E, T>& graph,
```
bobluppes, line 13: Let's also consider changing the return type here. Can we come up with a return type which is both:

- Efficient in memory layout
- Easy and intuitive to use (this is the main concern of the library)

include/graaflib/algorithm/shortest_path/dijkstra_shortest_paths.tpp
bobluppes: Trimmed, thanks.

include/graaflib/algorithm/shortest_path/dijkstra_shortest_paths.tpp
bobluppes: Trimmed, thanks.

include/graaflib/algorithm/shortest_path/dijkstra_shortest_paths.tpp
bobluppes: Filed https://github.com/bobluppes/graaf/issues/458 to track this - no action taken here.

include/graaflib/algorithm/shortest_path/dijkstra_shortest_paths.tpp
bobluppes: Filed https://github.com/bobluppes/graaf/issues/459 to scope this properly (outer unordered_map + inner std::list, plus bellman_ford which shares the same return shape). Related to #458 since a shared dense-storage utility could double as the new return container. No change made in this PR.

PR #455, opened 2026-09-20
perf: back greedy_graph_coloring's per-edge lookups with a vector

bobluppes:
## Summary
Sixth of a staged series for #443 (BFS in #449, Kruskal in #451, Prim in #452, `dijkstra_shortest_path` in #453, `dijkstra_shortest_paths` in #454).

The function's return type (`unordered_map<vertex_id_t, int>`) is a documented public contract, so it's unchanged - same situation as `dijkstra_shortest_paths` in #454. The hot path is `consider_neighbor()`, called once per (neighbor, predecessor) pair for every vertex - i.e. O(E) times total - which looked up the in-progress coloring by hash map. That's now a dense vector, indexed directly by id and grown lazily; the public map is only built once, in a single O(V) pass at the end.

`predecessors` (from the shared `get_predecessors()` utility) is unchanged: it's looked up once per *vertex*, not once per *edge*, so it isn't the hot path this issue targets.

## Benchmarks (Release, `perf/graaflib/greedy_graph_coloring_benchmark.cpp`, real datasets)

|---|---|---|---|
| web_google | 2159.6 ms | 1410.7 ms | 1.53x |
| web_berkstan | 1068.9 ms | 889.8 ms | 1.20x |

- web_google: 3.25x -> 2.13x
- web_berkstan: 9.25x -> 7.70x

## Test plan
- [x] `Graaf_perf`: `greedy_graph_coloring_benchmark` run in Release, before vs. after, see table above


bobluppes:
> The function's return type (unordered_map<vertex_id_t, int>) is a documented public contract, so it's unchanged

Let's consider some alternatives here. Given that we will in the near future bump the major version we have some freedom to change the public contract. Can we come up with a return type which is both:

- Efficient in memory layout
- Easy and intuitive to use (this is the main concern of the library)

include/graaflib/algorithm/coloring/greedy_graph_coloring.tpp
```diff
@@ -24,16 +23,30 @@ std::unordered_map<vertex_id_t, int> greedy_graph_coloring(const GRAPH& graph) {
     predecessors = get_predecessors(graph);
   }
 
+  // Indexed directly by vertex_id_t rather than an unordered_map for the
```
bobluppes: Also here, don't mention what you did not do (unordered_map).

bobluppes: Trimmed, thanks.

PR #456, opened 2026-09-20
perf: back welsh_powell's per-edge color lookups with a vector

bobluppes:
## Summary
Seventh of a staged series for #443 (BFS in #449, Kruskal in #451, Prim in #452, `dijkstra_shortest_path` in #453, `dijkstra_shortest_paths` in #454, `greedy_graph_coloring` in #455).

Same shape as #455: the function's return type (`unordered_map<vertex_id_t, int>`) is a documented public contract, so it's unchanged. `collect_neighbor_color()` - called once per (neighbor, predecessor) pair for every vertex, i.e. O(E) times total - looked up the in-progress coloring by hash map. That's now a dense vector, indexed directly by id and grown lazily; the public map is only built once, in a single O(V) pass at the end.

`neighbor_colors` (the set of colors already used by a vertex's neighbors, used to find the smallest free color) stays an `unordered_set<int>`: it's keyed by *color*, not `vertex_id_t`, so it's out of scope for this issue.

## Benchmarks (Release, `perf/graaflib/welsh_powell_benchmark.cpp`, real datasets)

|---|---|---|---|
| web_google | 4062.6 ms | 2270.5 ms | 1.79x |
| web_berkstan | 2918.6 ms | 1782.3 ms | 1.64x |

- web_google: 4.64x -> 2.59x
- web_berkstan: 9.35x -> 5.71x

## Test plan
- [x] `Graaf_perf`: `welsh_powell_benchmark` run in Release, before vs. after, see table above


bobluppes:
> Same shape as https://github.com/bobluppes/graaf/pull/455: the function's return type (unordered_map<vertex_id_t, int>) is a documented public contract, so it's unchanged.

Let's consider some alternatives here. Given that we will in the near future bump the major version we have some freedom to change the public contract. Can we come up with a return type which is both:

    Efficient in memory layout
    Easy and intuitive to use (this is the main concern of the library)

include/graaflib/algorithm/coloring/welsh_powell.tpp
```diff
@@ -39,8 +40,21 @@ std::unordered_map<vertex_id_t, int> welsh_powell_coloring(const GRAPH& graph) {
 
   std::sort(degree_vertex_pairs.rbegin(), degree_vertex_pairs.rend());
 
-  // Step 2: Assign colors to vertices
-  std::unordered_map<vertex_id_t, int> color_map;
+  // Step 2: Assign colors to vertices.
+  //
+  // Indexed directly by vertex_id_t rather than an unordered_map for the
```
bobluppes: Also here, don't mention what you did not do (unordered_map).

bobluppes: Trimmed, thanks.

